PR #48, opened 2017-12-18
Refresh ephemera/ files

spiffxp:
They were originally copied from kubernetes/community/contributors/devel/release and then never tombstoned or deprecated over there, so they caught updates.

If this repo makes sense as the authoritative source for this content, I'd like to tombstone the copies in kubernetes/community to point to here.

I also thought @enisoc's patch manager playbook was useful so moved it someplace more permanent sounding.


jdumars:
Wow!!!

PR #55, opened 2017-12-22
Removing "Secondary release manager" role (dup by the Lead Shadow)

idvoretskyi:
Removing the "Secondary release manager" role as it's duplicated by the Release Lead Shadow.

PR #56, opened 2017-12-22
Release Lead Secondary renamed to the Release Lead Shadow

idvoretskyi:
Fixes https://github.com/kubernetes/sig-release/pull/55#issuecomment-353639955


README.md
```diff
@@ -154,9 +154,8 @@ All the roles defined below report to the Release Management Team Lead.
   and all deadlines leading up to release
 - Manages [exception](https://github.com/kubernetes/features/blob/master/EXCEPTIONS.md) process for features that want to merge after code freeze
 
-### Release Team Secondary
-The Release Team Secondary is an assistant to the Release Team Lead and works to fill gaps in Release Team staffing or effort. The Secondary should
-be familiar with the release process and remain ready to discharge the responsibilities of the Lead in the event the Lead is unavailable.
+### Release Team Shadow
```
enisoc: Technically this should be "Release Team Lead Shadow" to distinguish it from "Release Team Shadow" below, and also to avoid having to fix the TOC link by adding `-1` to the anchor name.

But maybe we don't even need this, if we combine it with the general shadow section below.

By the way, below we say the release team member should select mentees, but in the last two cycles we just let people volunteer to shadow. Should we update the wording to match what we actually do?

idvoretskyi: > Should we update the wording to match what we actually do?

Agree.

idvoretskyi: @enisoc updated.

README.md
```diff
@@ -10,15 +10,15 @@
...
+   - [Release Team Shadow](#release-team-shadow)
    - [Release Branch Manager](#release-branch-manager)
    - [Docs Lead](#docs-lead)
    - [Features Lead](#features-lead)
    - [Bug Triage Lead](#bug-triage-lead)
    - [Test Infra Lead](#test-infra-lead)
    - [Upgrade Testing Lead](#upgrade-testing-lead)
    - [CI Signal Lead](#ci-signal-lead)
-   - [Release Team Shadow](#release-team-shadow)
+   - [Release Team Lead Shadow](#release-team-lead-shadow)
```
enisoc: This TOC entry should be swapped with the one you changed above.

idvoretskyi: @enisoc oooops, fixed.
Thanks!

PR #61, opened 2018-01-02
Change from "marketing" to "communications"

jdumars:
In the process of documenting this role (PR soon!), it became clear that this is not a "marketing" role, but instead a communications coordinator position between CNCF, media, PM, and the release team.  There is an assumption that the CNCF will have a person dedicated to their marketing efforts already.


jdumars:
Role draft: https://docs.google.com/document/d/11uwwCUoZxoyhtD1_HkVmEvDooq9VZ6F5sy_3izaKEtY

Will be a PR once reviewed by CNCF.

idvoretskyi:
@jdumars +1, it's a more clear definition of the role.

PR #65, opened 2018-01-04
Add OWNERS for PST docs

cblecker:
Adds OWNERS file for PST documentation, and github usernames to the docs itself.

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -23,3 +23,9 @@ aliases:
     - roberthbailey
     - saad-ali
     - wojtek-t
+  product-security-team:
+    - philips
+    - jessfraz
+    - cjcullen
+    - tallclair
+    - liggitt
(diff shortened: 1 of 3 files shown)


spiffxp:
hmm, let's see if this merges, github's ui is telling me we'd be applying the `needs-rebase` label if we had that turned on here

cblecker:
@spiffxp it's because "Require branches to be up to date before merging" is enabled in the branch protection options for this repo. Typically, it's off.
ref: https://help.github.com/articles/enabling-required-status-checks/

spiffxp:
I guess the real question is whether this is the sort of thing we'd want tide to pay attention to, for those repos that would prefer to keep using that option

cblecker:
My understanding is that it's so testing can be done just on the branch, but ensuring that those tests also include all the commits from master. However, as prow handles merging the branch on top of master before testing anyways, it shouldn't apply and just presents an odd UI. In the PR that @fejta has cooking for a prow branch protection plugin (https://github.com/kubernetes/test-infra/pull/6077), he is blindly setting it to false (I assume for this reason).

PR #98, opened 2018-03-27
Translate Google doc burndown meeting into MD

jdumars:
This is the complete meeting history for the 1.10 release. Some future generation may want to do technoarchaeology to understand the origin of kubosapiens.


jdumars:
There definitely formatting errors, but I will die of old age before I am able to fix them all. Perfect is the enemy of done in this case.

idvoretskyi:
@jdumars 

> There definitely formatting errors, but I will die of old age before I am able to fix them all. Perfect is the enemy of done in this case.

Long live, kubosapiens!

PR #100, opened 2018-03-28
Add mostly empty release team directory for 1.11

calebamiles:
...but note that Josh is our Release Team Lead and Ihor will likely lead features again. The idea is to merge this skeleton quickly and then add the rest of the release info as it's available


jberkus:
LGTM, will have more team details and draft schedule tomorrow.

/lgtm

idvoretskyi:
> Ihor will likely lead features again.

Yaaaa, but finally I would like to have a shadow!

jberkus:
I'll put out a call, but leads also need to try to recruit shadows.

PR #108, opened 2018-03-29
v1.11 release team - add colemickens as shadow bug triage

colemickens:
Hello. I'd like to volunteer to do `Bug Triage` for the v1.11 release cycle. (I skimmed the README, but didn't see a formal process for this, hopefully this is sufficient.)

PR #111, opened 2018-04-03
Add Cole Mickens as shadow bug triage.

jberkus:


colemickens:
Hi @jberkus, this looks like the same change as #108, which is passing CLA, etc, now. Thanks.

jberkus:
@colemickens aha, you revised that one!  Closing this one.

PR #123, opened 2018-04-09
Update CI Signal role in README.md

AishSundar:
From my understanding and past history, the CI Signal lead also monitors and tracks issues with the Upgrade / downgrade tests as well. Modifying the README to reflect this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,8 +200,8 @@ be familiar with the release process and remain ready to discharge the responsib
   finding manual testing volunteers, and ensuring any issues discovered are communicated widely and fixed quickly
 
 ### CI Signal Lead
-- Ensures that all non-upgrade test CI provides a clear go/no-go signal for the release
-- Tracks and finds owners to fix any issues with any (non-upgrade) tests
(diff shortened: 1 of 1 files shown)


AishSundar:
@spiffxp - Thanks for pointing me to the Upgrade lead notes. Deleted the section. 

While I plan to cover the first 3 points that were listed under Upgrade lead role, I am not sure if CI Signal lead should support the following:

- Organizes the manual upgrade testing efforts, including setting up instructions for manual testing,
-  finding manual testing volunteers, and ensuring any issues discovered are communicated widely and fixed quickly.

If you feel its comes under the Signal lead jurisdiction I am happy to add, but will need support from others in RT to get it going.

spiffxp:
I'm in favor of dropping manual upgrade testing entirely, it's something we haven't done for a while AFAIK (I know it wasn't happening while I was 1.9 CI signal lead)

If we find ourselves in a position where manual testing needs to happen, we've probably got bigger problems, so I imagine the team as a whole would be rallying around the effort

PR #173, opened 2018-06-08
Added major themes section

nickchase:


calebamiles:
@nickchase I would merge this skeleton in and then have SIGs fill out their own sections in separate PRs

PR #177, opened 2018-06-11
[1.11] Mark Heapster as deprecated

DirectXMan12:
The Heapster project has been deprecated, so we should warn about that
in the Kubernetes 1.11 release notes.

### DIFF
--- a/releases/release-1.11/release_notes_draft.md
+++ b/releases/release-1.11/release_notes_draft.md
@@ -68,6 +68,7 @@ This tag is for changes that don't affect the user experience, such as behind-th
 
 ### sig-instrumentation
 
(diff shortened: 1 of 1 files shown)


DirectXMan12:
cc @kubernetes/sig-instrumentation-misc 
@brancz do we want to add this to "Major Themes" too?

brancz:
> do we want to add this to "Major Themes" too?

Yes I think it's important to make this as widely known as possible.

DirectXMan12:
I'll update the PR to add it there too, then.

DirectXMan12:
It's now in the major themes section as well.  PTAL

brancz:
Thanks for taking care of this!

/lgtm

PR #181, opened 2018-06-15
update 1.11 release note draft with sig-auth highlights

liggitt:


justaugustus:
@tpepper @nickchase -- what do we want to do with these?

liggitt:
It was already merged in the release note draft

PR #194, opened 2018-06-20
Update IPVS topics in release notes draft

Lion-Wei:


Lion-Wei:
@m1093782566 @calebamiles PTAL, thanks.

jhorwit2:
@Lion-Wei should we also call out the known issue with graceful termination here? https://github.com/kubernetes/kubernetes/issues/57841

calebamiles:
/approve
/lgtm
/hold

@Lion-Wei please resolve the comment from @jhorwit2 and remove the hold. Thanks!

Lion-Wei:
@calebamiles Thanks.
@jhorwit2 , Thanks for reminding. We do have this issue in discuss, but I'm not sure is this necessary to call out a known issue.
@m1093782566 , what do you think?

m1093782566:
what about creating an action-required item？

Lion-Wei:
I think known issue is fine. 
@calebamiles , Should I add known issue to `release_notes_draft.md` or other document? Cause I didn't found any known issue related item in `release_notes_draft.md`.
Thanks.

Lion-Wei:
I think the known issue description should be like: 
```
Known Issue

IPVS based kube-proxy doesn't support graceful close connections for terminating pod. This issue will be fixed in a future release.
For more information, see https://github.com/kubernetes/kubernetes/issues/57841.
```
WHat do you think? @jhorwit2

PR #199, opened 2018-06-21
1.12: add to branch manager lead/shadow

tpepper:
I've got two more volunteers on branch management:  Doug MacEachern
to lead, and Travis Rhoden to shadow.  The set of four shadows on
branch manager should generate a good set of questions and along
with Caleb Miles' help get process documentation created for the
role and a solid pipeline for future leads on role.  Also these
folks should be good candidates to shadow and eventually lead in
the patch release manager role as their short term experience on
release branch management can translate there.


calebamiles:
🥇, thanks @tpepper! 
/lgtm

PR #214, opened 2018-07-02
1.12 Release Notes so far (new tool format proposal)

marpaia:
I generated these notes using the new `release-notes` tool that I'm working on in https://github.com/kubernetes/release/pull/575. I would like to just continuously re-render these notes until it comes time to give them to the individual SIGs as the release burns down. So if there are any edits that we'd like to structurally incorporate into this document, let's encode it in the code so that we can keep re-applying the same patterns for notes sorting (if possible).

For each note, we have the following information available to use for document organization decision making:

- The note text itself
- All the commit info (SHA, author, etc)
- Labels on the PR (`kind/bug`, `release-note-action-required`, `sig-node`, etc)
- The PR description and comments

Right now, I'm using all the SIG labels I can fine to pre-sort and then using the `kind/bug` and `release-note-action-required` label to add the "Bug Fixes" and "Action Required" sections respectively. Everything that doesn't have a SIG label or one of those two labels goes into "Other Notable Changes".

/assign @nickchase


jberkus:
additional labels to consider, that are sometimes present:

both of the below generally mean an API change or new API, and are likely to be present in addition to other kind/ labels:
kind/api-change
kind/design

Also, the kind/cleanup (or kind/refactor) label is occasionally present for PRs demanding a release note; do we want to do anything special with that?

marpaia:
Maybe `kind/feature` for a "New Features" section?

marpaia:
I added a "New Features" and "API Changes" section but there are none of those so far in 1.12. To get a good feel for what this would look like at the end of a release, take a look at this [rendering of the 1.11 release notes](https://gist.github.com/marpaia/acfdb889f362195bb683e9e09ce196bc).

Ideally "Other Notable Changes" wouldn't have many notes in it and many of these notes are in-fact associated with a SIG, but are just not labeled as such. Fortunately we can go label those PRs post-merge and re-render the release notes to get them sorted, but ideally we could require there is some categorization label applied to PRs which have a release note (ie: some `sig/*` or `kind/*` or `area/*` label).

marpaia:
I ran some experiments and from 1.10.0 to 1.11.0, the only PRs with release notes that didn't specify a SIG but did specify a kind label other than (and not including) one that would already categorize them (bug, feature, api-change, new-api) are:

- kind/old-docs for https://github.com/kubernetes/kubernetes/pull/64034
- kind/cleanup for https://github.com/kubernetes/kubernetes/pull/58784

So I think sig label and then kind/bug, kind/feature, kind/new-api, and kind/api-change make for a pretty comprehensive set of what is labeled at all. Past that, we can just generate the notes, see what's not labeled, go label it, and regenerate the notes.

marpaia:
Can I LGTM my own PR?

/lgtm

nickchase:
OK, I really like what we've got here.  One change that I think we need, though, is that "action required"  items shouldn't also appear in the sig sections.  Other than that, I think this is ready for us to move forward.

marpaia:
@calebamiles what do you mean? do you want the Action Required section subcategorized by SIG or anything in particular?

PR #230, opened 2018-07-24
Deletes cherrypicks-md file and uses k/community's cherry-picks.md as…

guineveresaenger:
… single source of truth

Cherry picks are a contributor issue, and we should not have duplicate files with differnet histories sitting around. All cherry-pick instructions should live in the contributors section of k/community.

Fixes https://github.com/kubernetes/community/issues/1760 (see discussion).


spiffxp:
/lgtm
/approve
Agree k/community is authoritative, and I cackle at glee with nuking something in ephemera

PR #232, opened 2018-07-24
Setup OWNERS files for each role handbook

spiffxp:
Followup to https://github.com/kubernetes/test-infra/pull/8798

When I'm updating the docs for a given release-team role, I want prow to auto-request reviews from people who currently or have previously staffed that role.  

I populated aliases based on 1.7 forward.

I would have used a regex-based OWNERS file, but that doesn't cooperate with the `approve` plugin right now (https://github.com/kubernetes/test-infra/issues/7690) so I had to breakup into dirs per role.

I searched org:kubernetes for references to the role handbooks and found none, so I haven't dropped tombstones, though I can be convinced if necessary.

WDYT?


justaugustus:
Looks great, @spiffxp!
Shouldn't be any need for tombstones as this directory re-org is pretty fresh as of https://github.com/kubernetes/sig-release/pull/216

/lgtm
/hold cancel

PR #264, opened 2018-08-25
Update RT Lead docs to remove references to `release-x.y.md`

justaugustus:
- Update RT Lead docs to remove references to `release-x.y.md`
- Add note about updating blockade config for previous releases

Follow-on to #263 
rel: https://kubernetes.slack.com/archives/C2C40FMNF/p1534957335000100

/assign @tpepper 
/cc @spiffxp

### DIFF
--- a/release-team/role-handbooks/release-team-lead/README.md
+++ b/release-team/role-handbooks/release-team-lead/README.md
@@ -60,9 +60,9 @@ Release Team selection should happen in accordance with the [Release Team Select
 
     * kubernetes/sig-release
 
-        * releases/release-x.x
+        * releases/release-x.y
 
-            * release-x.x.md (release schedule)
+            * README.md (release schedule)
 
             * release-notes-draft.md (consumed by the automated release process)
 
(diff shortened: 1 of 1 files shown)


tpepper:
Do we actually need to move these or conceptually can we just stop editing them and treat them as archived in place?

The "Update RT Lead docs to remove references to `release-x.y.md`" patch looks good.

spiffxp:
Agree with @tpepper not sure I see a need to archive.

If you want to prevent PR's to old release docs, you could setup a blockade config.  See for example the [config we use to prevent PRs to old events in k/community](https://github.com/kubernetes/test-infra/blob/master/prow/plugins.yaml#L126-L131)

justaugustus:
I've updated this to remove the archives and instead make mention of updating blockade to prevent pushes to all but the most recent three release folders.

Current aPR to add blockade config is here: https://github.com/kubernetes/test-infra/pull/9250

justaugustus:
Holding until https://github.com/kubernetes/sig-release/pull/271 is merged.
/hold

justaugustus:
This is ready to go now.
/hold cancel

PR #265, opened 2018-08-25
Add `sig/release` label to OWNERS

justaugustus:
Everything in this repo is governed under SIG Release, issues / PRs such be tagged as such.
This adds automatically labels issues / PRs opened with `sig/release`

/sig release

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -5,3 +5,5 @@ approvers:
   - jdumars
   - sig-release-leads
   - release-team-lead-role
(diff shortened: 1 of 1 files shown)


spiffxp:
Note: this isn't going to work unless the `owners-label` plugin is turned on for this repo.

See https://github.com/kubernetes/test-infra/pull/9011 for an example of me doing this for kubernetes/test-infra

Issue #233, opened 2018-07-25
Archive pre-1.12 Features Tracking spreadsheets


justaugustus:
Suggestion from @calebamiles: https://github.com/kubernetes/sig-release/pull/231#issuecomment-407727489

/sig pm
/sig release
/assign

PR #231, opened 2018-07-24
[RFC] Add Features Lead Handbook

justaugustus:
Documenting the Features Lead process.
WIP, but I'm getting this out of a [Google Doc](https://docs.google.com/document/d/1WPB0N4kjVN_CY205dBEzs9KULWgkBeW1GASV4YHaK_w/edit?usp=sharing) and up for review. :)

Closes: #124 

/sig release pm
/cc @idvoretskyi @jdumars @jberkus @tpepper 
/hold

PR #270, opened 2018-09-04
Add features archives for past releases (1.3 - 1.11)

justaugustus:
Per Caleb's suggestion in https://github.com/kubernetes/sig-release/pull/231#issuecomment-407727489:
> ... I would also suggest exporting the features tracking spreadsheet as a CSV for archival in the release-X.Y directory as one of the tasks but otherwise looks good to me. Thanks!

Closes: https://github.com/kubernetes/sig-release/issues/233

/assign @calebamiles @tpepper 
/sig pm
/sig release

Signed-off-by: Stephen Augustus <foo@agst.us>


idvoretskyi:
We've been maintaining the features spreadsheets since 1.3 release. Why have we started with 1.9 in this PR?

justaugustus:
@idvoretskyi -- good point, I can add those too.

justaugustus:
@idvoretskyi -- I've update this to include features going back to 1.3.

Looks like the blockade (https://github.com/kubernetes/test-infra/pull/9250) I put in worked as well (https://github.com/kubernetes/sig-release/pull/270#issuecomment-418539177)!

Let's see if we can override that...
/override blockade

justaugustus:
@idvoretskyi @calebamiles -- I'm not exactly sure, but I think someone with manual merge powers would have to get this in (to get past the blockade).

PR #278, opened 2018-09-05
tombstone old release information (subprojects)

calebamiles:
...affects `release-1.{3,4,5,6}/` directories


spiffxp:
I'm not convinced.  What's the case for removing the older release dirs?  This is one of the few times where "it's a git repo, we can always restore it" doesn't make sense to me.

calebamiles:
Well I would rather not enable `blockade` for the `sig-release` repo in order to "archive" these directories so I would like to either keep all the directories as is (perhaps with an `OWNERS` file that only has the SIG leads in it in order to gate changes to the older releases). If we're fine just rolling back to where we were a few days ago I'm happy to close this

PR #298, opened 2018-09-16
Rename Features to Enhancements

justaugustus:
/sig release
/assign @kacole2 @AishSundar @tpepper 

Signed-off-by: Stephen Augustus <foo@agst.us>

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -33,9 +33,10 @@ aliases:
     - krzyzacy
     - aishsundar
     - mohammedzee1000
-  features-role:
+  enhancements-role:
     - idvoretskyi
     - justaugustus
+    - kacole2
   test-infra-role:
     - krzyzacy
     - bentheelder
--- a/release-team/role-handbooks/enhancements/OWNERS
+++ b/release-team/role-handbooks/enhancements/OWNERS
@@ -1,0 +1,4 @@
+approvers:
+  - enhancements-role
+reviewers:
+  - enhancements-role
--- a/release-team/role-handbooks/enhancements/README.md
+++ b/release-team/role-handbooks/enhancements/README.md
@@ -1,39 +1,39 @@
-# Features Lead Handbook
+# Enhancements Lead Handbook
 
 ## Overview
 
-While the Features Lead serves as a member of the Release Team (a subproject of [SIG Release][sig-release]), this role is also a liaison to [SIG PM][sig-pm] across the following subprojects:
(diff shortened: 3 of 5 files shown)


justaugustus:
@AishSundar -- I can search and replace the relevant role handbooks once this merges.

justaugustus:
@idvoretskyi -- it was mentioned during the [Release Team meeting on 9/10](https://youtu.be/ASHjFacX8t4?t=28m25s), during our 1.13 nominations, to no objections. 

Fair point though, should we announce this wider? The idea here was to essentially rebrand the role to fall more in line with the fact that we need to be tracking enhancements (KEPs, features, anything that falls under that umbrella) as opposed to just Features, as the KEPs process evolves.

Let me know what you think.

nikhita:
> Fair point though, should we announce this wider? The idea here was to essentially rebrand the role to fall more in line with the fact that we need to be tracking enhancements (KEPs, features, anything that falls under that umbrella) as opposed to just Features, as the KEPs process evolves.

Slightly related: should the label `kind/feature` be relabelled as `kind/enhancement` (or be introduced as a new label)?

/cc @spiffxp

justaugustus:
@nikhita -- I'm not sure exactly where it was mentioned, but I remember there was discussion around deprecating `kind/enhancement`. It was subsequently replaced with `kind/feature`.

Enhancements ~= KEPs + Features

From https://github.com/kubernetes/community/pull/2655#discussion_r218637790:
> We're introducing the idea of an umbrella term: `Enhancements`.
> Enhancements are anything that modifies or adds new functionality to Kubernetes.
> 
> So KEPs and Features would both fall under that category.
> 
> KEP: multi-release event
> Feature: hopefully only a single-release event
> 
> Features should roll into a KEP.
> 
> I think a good current example might be CoreDNS.
> 
> * Implementing CoreDNS == KEP
> * CoreDNS as default in kubeadm == Feature

That said, we're working on [introducing `kind/kep`](https://github.com/kubernetes/test-infra/pull/9437), which would cover tracking KEPs. No need for an additional `kind/enhancement` label.

idvoretskyi:
@justaugustus 

> Fair point though, should we announce this wider? 

It would be great. My concern was not about the renaming itself (I'm definitely fine with that!), but about the agreement that hasn't been shared publicly.

idvoretskyi:
@justaugustus feel free to cancel `hold` once you'll be ready, looks good to me after the explanation.

AishSundar:
Cool, now that this is in I will make the announcement as part of 1.13 intro mail as well !

justaugustus:
@AishSundar go for it! :tada:

PR #302, opened 2018-09-19
Request that we publish Debs and RPMs for beta and release candidates

dims:
Change-Id: I3eac113f7ad23b0e813d9fb59fbd9241887da9ec

### DIFF
--- a/release-team/role-handbooks/branch-manager/README.md
+++ b/release-team/role-handbooks/branch-manager/README.md
@@ -113,7 +113,10 @@ Adding the --rc flag switches behavior on to building release candidates.  Again
 * `./gcbmgr stage release-1.12 --official --build-at-head --nomock`
 
 ## Debs and RPMs
-These require an additional layer of build and publish, which is currently still done by a Googler.  After building, but before release notification, ping @mbohlool on slack indicating for which build (for example: 1.12.0-beta.1) Debs and RPMs are required.
+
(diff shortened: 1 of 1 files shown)


neolit123:
+1

calebamiles:
/hold

@dims I don't believe this is actionable at this point. In my opinion we need to

1. blockade all deb/rpm specifications which live here in favor of those which live in [kubernetes/build][] (this also solves a large part of the "release tooling should be versioned" problem as the specs will be bundled with the version of Kubernetes that they apply to)
1. migrate deb/rpm hosting to a community maintained project (rather than being hosted on Google repositories)
1. move deb/rpm generation into `anago` so that the process actually becomes automated and stops lagging image publishing
1. profit

Rather than creating [issues][] in a poorly trafficed repository this really should be worked as part of a KEP given the impact. 

**cc: @jagosan**

[kubernetes/build]: https://github.com/kubernetes/kubernetes/tree/master/build
[issues]: https://github.com/kubernetes/release/issues/631

dims:
@calebamiles ack on all the points :) @spiffxp told me to!

hoegaarden:
In addidtion to Caleb's point:

> 2. migrate deb/rpm hosting to a community maintained project (rather than being hosted on Google repositories)

If we publish RC debs, we should publish them as a separate distribution (compare: debian's `stable`, `testing`, `unstable`, `experimental`). Then users can opt-in to the RC releases (by adding this new distribution and setup pinning or explicitly install from that distribution via `apt-get install -t ...` or such) and won't be surprised by an RC release coming in all of a sudden.
Same should be done for rpms.

PR #308, opened 2018-09-24
security-release-process: require certified Kubernetes distributions

philips:
Certification creates a clean cut decision process for the PST to decide
whether a person or organization can join the list. It beats the
alternative which is the PST trying to make a judgement call on their
own.

Further clarify that you will be removed from the list if your project
stops meeting criteria and encourage use of email aliases.


philips:
cc @cjcullen @jessfraz @tallclair @liggitt

PR #311, opened 2018-09-26
Clarify bug bar for information disclosure & lower severity of some cases

tallclair:
It's hard to cover every possible scenario, but in the case of non-identifying information disclosure where it is hard to imagine what an attacker would do with the data, I do not think we should consider that a High vulnerability.

/cc @cjcullen @jessfraz @philips @liggitt


jessfraz:
nice one lgtm :)

i like that as we grow this will evolve as well, its kinda awesome

PR #318, opened 2018-09-28
Adjusting CI signal shadow membership, since I can't reasonably handl…

jberkus:
…e more than two shadows.  Everyone has been informed.

Aish and I met about this today and after a long discussion picked the two shadows, which for 1.13 will be Mortent & Maria.


AishSundar:
Thanks for your diligence in the shadow selection process @jberkus 

/lgtm

PR #325, opened 2018-10-04
Add a RC2 date on 1.13 release calendar

AishSundar:
Following up on the comments from 1.12 Retro to have 2 RC builds during the release cycle. I thought we had in on the schedule, but looks like I confused this with the 2 Beta cuts.

/assign @dims 

/cc @spiffxp @dougm


dims:
/lgtm
/hold

please cancel hold when appropriate. (hold is for extra review/eyes)

PR #330, opened 2018-10-07
Add bit.ly/k8s113-bugtriage

nikopen:
prototype automation of bug triage sheet using googlesheet functions


AishSundar:
Thanks for doing this.

/lgtm

Issue #281, opened 2018-09-06
Clean up old release OWNERS


justaugustus:
As we move into new release cycles, we should also seek to update OWNERS across the repo to only include release team lead roles from the past, say, N - 3 or N - 4 cycles.

There should also be instructions about this written in the Release Team Lead handbook.

/sig release
/good-first-issue
cc: @calebamiles @idvoretskyi

idvoretskyi:
@justaugustus makes sense, +1!

justaugustus:
Fixed in #341.
/remove help
/assign

PR #341, opened 2018-10-16
Update OWNERS_ALIASES to only include N-3 RT personnel

justaugustus:
Fixes: #281 

/sig release
/assign @calebamiles @spiffxp @AishSundar @tpepper 

(let me now if the comments break automation and I'll remove them)

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -3,72 +3,47 @@ aliases:
     - calebamiles
     - jdumars
   release-team-lead-role:
-    - bgrant0607
-    - dchen1107
-    - enisoc
-    - erictune
-    - ethernetdan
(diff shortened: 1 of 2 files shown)


AishSundar:
One nit, otherwise. Holdng for others to review

/lgtm
/hold

spiffxp:
/lgtm
Though I do wish we had a better way of giving emeritus recognition, dropping people from the review pool is at least a bit of a treat.

justaugustus:
@spiffxp -- yeah, we have them listed in the `release-x.y/release_team.md` files, but agreed that there should be a more visible way of recognizing those contributions.

If anyone thinks of something, I can PR it later, but in the meantime...
/hold cancel

PR #365, opened 2018-11-09
Copy editing the Release Notes role handbook

marpaia:
This PR contains a few copy-edits and markdown formatting changes to the Release Notes Handbook.

PR #366, opened 2018-11-10
Fix typos: privleges ->privileges

mooncak:
Fix typos: privleges ->privileges


marpaia:
The LoC with this typo is no longer in the handbook in my pending PR (#365). I think we should close this and just merge that.

mooncak:
> #365

ok, thanks.

PR #384, opened 2018-11-26
Add note about externalName behavior change in kubedns/coredns

chrisohaver:
CoreDNS and kube-dns differ in the way they handle Services with externalNames that _look like_ IP addresses.  CoreDNS follows the DNS spec here, whereas kube-dns does not.
This results in an apparent behavior change when moving from kube-dns to coredns.

More detail here: coredns/coredns#2324


AishSundar:
@chrisohaver the current working copy of the Release notes draft is https://docs.google.com/document/d/1fL_xUwEWbxdRen-2mljio-Yd8Z2fzh2cQjaIOHRSHG0/edit

Can you please add your note to that versus updating the draft here? @marpaia to guide in this regard.

chrisohaver:
Thanks for pointing me in the right direction!  I have suggested the note to the google doc.

Issue #372, opened 2018-11-15
Assemble Kubernetes 1.14 Release Team


justaugustus:
This is the official issue for assembling the Kubernetes 1.14 Release Team.

Please review the [Release Team Selection](https://git.k8s.io/sig-release/release-team/README.md#release-team-selection) docs for details on the process and criteria.

We are currently in discussions around improving / refining the selection process (https://github.com/kubernetes/sig-release/issues/368, https://github.com/kubernetes/sig-release/issues/369), but would like to open this issue to get the ball rolling.

/sig release
cc: @AishSundar @spiffxp @calebamiles @tpepper @jdumars

justaugustus:
So far, the following people have reached out to me to participate:
- @d-nishi - CI Signal shadow
- @nikhita - Patch Release Management Team
- @yue9944882 - Patch Release Management Team
- @lledru - Bug Triage shadow

jeefy:
I'd like to volunteer for Release Notes shadow! :)

idealhack:
I'd like to join the Patch Release Team.

Since I'm the Branch Manager Shadow in 1.12 and 1.13, I can continue to work with branch management if needed.

xmudrii:
Hey everyone! :wave: I'd love to join the 1.14 release team!

So far seeing by roles, if there is free space, I would love to be shadow for Patch Release or Branch Manager.

marpaia:
I'd like to nominate @dstrebel to be the 1.14 Release Notes Lead!

shubheksha:
I'd like to shadow bug triage again or CI signal!

christianh814:
I'd like to shadow someone

hhoover:
I’d like to shadow on the communications team

sumitranr:
I would like to shadow the patch release management

AishSundar:
I would like to officially nominate @spiffxp as the 1.14 Release lead :) Go SIG-Beard !!!

justaugustus:
SIG BEARD 4EVR!
/meow space

kacole2:
I would like to nominate @claurence as Enhancements Lead for 1.14. I will take a step back into a shadow/mentor role for 1.14 and take the opportunity to update some of the Enhancement Lead timetables/processes.

munnerz:
If there's more needed/space, I'd like to volunteer as a patch release shadow too 😄

tan-c:
I would like to shadow for CI signal team if there is still space :))

[...]

Issue #369, opened 2018-11-14
Define roles and processes for Patch Release Management Team


nikhita:
The release team is moving from the role of a Patch Release Manager to a Patch Release Management _Team_: https://github.com/kubernetes/sig-release/tree/master/release-team#kubernetes-release-team-roles.

The [current handbook](https://github.com/kubernetes/sig-release/blob/master/release-team/role-handbooks/patch-release-manager/README.md) defines processes and suggestions for a patch release manager.

We need to now define processes for the Patch Release Management Team instead. This can include (but not limited to) how many roles this team should have, what responsibilities these roles would entail, etc.

It is important to do this soon, given that we would starting the selection process for the 1.14 release team (https://github.com/kubernetes/sig-release/issues/368).

/assign @justaugustus @tpepper @aleksandra-malinowska

nikhita:
/kind design
/priority important-soon

cc @spiffxp @jberkus @calebamiles

justaugustus:
As an aside the Patch Release Team and Branch Manager roles in the Release Team will become extensions of the to-be-formed Release Engineering subproject within SIG Release.

yue9944882:
how about linking this document to patch team introduction too? highlighting that patch releases are meant for fixing production-impacting bugs 

[https://github.com/kubernetes/community/blob/master/contributors/design-proposals/release/versioning.md#patch-releases](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/release/versioning.md#patch-releases)

ttousai:
To summarize #408. Patch release manager should be renamed patch release management team for  1.14.

cc @spiffxp @tpepper @aleksandra-malinowska

tpepper:
We need to make it explicit what is the holiday coverage plan.  Traditionally the project slows down dramatically during major holidays.  Folks should know they can continue to trust in patch managers being appropriately reachable during these periods, especially the security team in the event of needing handling on critical vulnerabilities.  This hasn't been a problem generally, but coverage expectations should be made more clear.

tpepper:
It's my intention to finish 1.14 with this well defined and with it better defined move forward spreading it into a team effort, so perfect to add to milestone.

aleksandra-malinowska:
@justaugustus is the discussion going on somewhere about Release Engineering subproject? If so, does it make this issue obsolete?

justaugustus:
[...]

Issue #368, opened 2018-11-14
Proposal: selection process for Shadows


jberkus:
**The Problem:**

RT Section Lead/Shadow is a mentorship relationship, but we have not been treating it as one.  Instead of having a process to select the most suitable mentees for a given cycle, we simply accept anyone who volunteers before a certain deadline, up to four shadows per lead.  This has resulted in leads having too many shadows for them to manage, and shadows not being mentored for the role they are shadowing.  Further, we have contributors volunteering to shadow without any clear idea of what they're committing to or what the requirements and expectations are.

**The Proposal:**

We should have a semi-formal selection process, which should work as follows:

1. shadow applicant applies for their shadow position using a specific questionnaire for the shadow role.  A draft template for these questionnaires follows.
2. the section lead and the RT Lead (or shadow) review the questionnaires and pick the candidates with the necessary time availability and other requirements for the role.
3. the section lead and the RT Lead have a brief chat with each qualified candidate (by slack, video, or other convenient mechanism) to select for mentorship match.
4. The section lead and RT lead pick the final Shadows for that RT.

As a prerequisite to this, it would also be good to establish guidelines on how many shadows are the maximum practical for a given role.  For example, CI signal can't really make use of more than 2 shadows, but Docs is capable of keeping up to 4 shadows busy.

This also begs the question of whether we should have a process for selecting section leads.  To date, one hasn't seemed necessary because, unlike shadows, we have a shortage rather than a surplus of volunteers, and generally all section leads are well-known from prior release teams.

## draft template for shadow questionnaire

You have asked to be a _role_ Shadow for the _ver_ Release Team. This questionnaire is to determine how well you match the position's requirements, and to give you an idea what's involved.  In cases where there are too many applicants for a particular role, it may also be used to decide who is selected this release cycle.  Please note that you are expected to be inexperienced in some areas, as Shadow is a mentored role.

1. The _role_ handbook [link] defines the tasks and responsibilities of the role for this release cycle.  Are you prepared to assist in fulfilling these duties?  Please read the full handbook before answering.
 
2. _role_ Shadow requires a substantial time commitment, especially
during _busy_period_ (_dates_). This can be more than _time requirement_, and may include attending video meetings at inconvenient times.  Is this time you can commit to?

[...]

PR #408, opened 2018-12-05
Add Stable Branch Management Team

ttousai:
This PR adds a Stable Branch Management Team (two leads and shadows) which replaces the Patch Release Manager role. 

This seems like a natural progression from the historically one patch manager zero shadows to two leads (a.k.a team) no shadows and now a team of lead + shadows.

cc @tpepper @aleksandra-malinowska

### DIFF
--- a/releases/release-1.14/release_team.md
+++ b/releases/release-1.14/release_team.md
@@ -9,4 +9,4 @@
 | Docs | Jim Angel ([@jimangel](https://github.com/jimangel)) |  |
 | Release Notes | Dave Strebel ([@dstrebel](https://github.com/dstrebel)) |  |
 | Communications | Natasha Woods ([@nwoods3](https://github.com/nwoods3)) |  Kaitlyn Barnard ([@kbarnard10](https://github.com/kbarnard10)) |
-| Patch Release Managers | | |
+| Stable Branch Management Team  | Aleksandra Malinowska ([@aleksandra-malinowska](https://github.com/aleksandra-malinowska)), Tim Pepper ([@tpepper](https://github.com/tpepper)) | Abubakr-Sadik Nii Nai Davis ([@ttousai](https://github.com/ttousai)) |


tpepper:
This is a natural progression imho.

In addition to @ttousai, we've also had participation interest from @castrojo , @sumitranr, @feiskyer, @idealhack .

tpepper:
We've talk about this transition and I like that Abubakr-Sadik is nudging us to commit to doing something concrete.  1.14.x patch management would effectively begin in April.  So that gives us time to operationalize this, assuming we commit to doing so now.

spiffxp:
I'm opposed to the name "stable branch" and would rather it stay "patch release management team"

ttousai:
I signed it

nikhita:
In case we are going with the "lead + shadow" route, just want to chime in that I'm interested in acting as a shadow as well. :)

/cc @yue9944882 
who was interested too

tpepper:
@spiffxp 
"What's in a name? That which we call a rose
By any other name would smell as sweet."

But seriously, naming is hard and that can be a conversation for another day and another PR and this one should focus on name += team.

spiffxp:
Thanks @tpepper.  I am _definitely_ thrilled to see patch release management become a team thing.  Maybe we want to decouple the patch release manager role from the Release Team entirely, and have the "Release Team" and the "Patch Release Team".  Historically I haven't seen any involvement from the patch release manager role during the initial release process.  But agree, scope should be += team

justaugustus:
Honestly, I'd like to also roll Branch Manager in as a role of the Patch Release Management Team.
Then the entire PRMT (oh no, an acronym) would act as an extension of the Release Engineering subproject (which yes, I know doesn't have a KEP yet) attached to the Release Team, similar to how Communications functions.

The reason for that being so that they will have visibility into the Release Team processes and are required to attend / be actively invested in the conversations that happen in the RT meetings.

I'd like to consolidate the conversations though, and will be closing this PR.
If you have thoughts on the PRMT, please pile on here: https://github.com/kubernetes/sig-release/issues/369

If you're interested in volunteering for a role on the 1.14 Release Team, please note that over here: https://github.com/kubernetes/sig-release/issues/372
We'll be working on collating the shadows and reaching out to them once the [questionnaire](https://github.com/kubernetes/sig-release/issues/368) is complete. Please be patient as we're also working through KubeCon preparations.

/close

PR #423, opened 2019-01-02
document cve request process

### LINKED ISSUE #282: document CNA process for CVE
via @liggitt 

we have CNA (CVE numbering authority) for kubernetes CVEs via https://github.com/distributedweaknessfiling/DWF-Documentation

bringing another person on board to issue CVEs for k8s would be done through that. since having more than one person involved adds coordination requirements around numbering, I wasn't planning to do that unless the volume grew, but if we want redundancy, we can pursue that

at the very least, documenting how CNA training/qualification works is important

liggitt:
/assign @philips 
fixes #282


philips:
thanks @liggitt !

Issue #427, opened 2019-01-03
Proposal: v1.13 patch release schedule


spiffxp:
/assign @tpepper @aleksandra-malinowska 
/kind documentation

I'd like to propose that the v1.13 patch release team lay out a schedule for patch releases.  I have often seen requests for "when is the X patch release getting cut" in many slack channels.  I suspect having this information ahead of time would save some angst.

aleksandra-malinowska:
Sounds good. Where should we put it?

tpepper:
We've got separate folders per release in sig-release repo, and the release team members are listed there.  I think as we shift toward a sustaining team of sorts, it would be good to give that its own unique landing page.  The individual release team markdown docs could link to that one page.

This would make it more discoverable for folks wanting to know what are current releases, current supported branches, when to expect next releases, and how to get code into a next release.  A portion of that is covered in https://github.com/kubernetes/community/blob/master/contributors/devel/release.md but again that's not a place folks are going to immediately see.

justaugustus:
@tpepper @aleksandra-malinowska -- are we in a good place to close this out?

justaugustus:
Closing this as the patch release schedule now exists: https://github.com/kubernetes/sig-release/blob/master/releases/patch-releases.md
/close

PR #429, opened 2019-01-04
Add 1.13 patch releases info

aleksandra-malinowska:
Quick draft w/ information about 1.13 patch release schedule proposed in #427. It's not meant to be perfect/include everything - I suspect we may have to iterate on it to find out what's most useful.

/kind documentation
/cc @tpepper @spiffxp

### DIFF
--- a/releases/patch-releases.md
+++ b/releases/patch-releases.md
@@ -1,0 +1,55 @@
+# Kubernetes patch releases' schedule
+
+Schedule and team contact information for Kubernetes patch releases.
+
+For general information about Kubernetes release cycle, see https://github.com/kubernetes/community/blob/master/contributors/devel/release.md
+
+# 1.13
+
+## Patch Release Management Team
+
+To discuss a cherry-pick, schedule, or anything else related to 1.13 patch releases,
+contact any member of the team. You can just tag us on #sig-release Slack channel
+or message directly. Please give us a business day to respond - we may be in a different timezone!
+
+Cherry-picks are monitored on regular basis, so it's likely you'll hear
+from us on PRs to release-1.13 branch by the time the next patch release is due.
+
+| **Github username** | **Slack username** |
+| --- | --- |
+| @aleksandra-malinowska | @aleksandram |
+| @tpepper | @tpepper |
+
+## Timeline
+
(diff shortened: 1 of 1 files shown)


releases/release-1.13/patch_releases.md
```diff
@@ -0,0 +1,37 @@
...
+branch have to merge to be included in the patch release. Any PRs left pending past
+this point will either wait for the next patch release, or may in rare cases
+delay current release.
+
+Target date is the date on which the patch release is cut. Must be at least 2
+days after cherry-picks deadline to allow all release-blocking tests to run.
+
+| **Patch Release** | **Cherry-picks deadline** | **Target date** | **Owner** |
+| --- | --- | --- | --- |
+| 1.13.2 | TBA | 2019-01-10 | @tpepper |
```
spiffxp: I'll leave it to you to decide what's appropriate, but I would personally prefer to see a longer forecast.  Something like a suggested cadence, or the next N potential dates, etc.

tpepper:
@feiskyer and @foxish what do you think about having one file, for example releases/patch_releases.md which has a section for each supported branch.  @aleksandra-malinowska 's proposed a 1.13 specific file and information.  I think it would be more easily discovered if the cadence and status info for all the branches was in one place.  Today's set of patch release managers should be able to commit to one file without much conflict.

aleksandra-malinowska:
@tpepper I wanted to add some schedule information instead of getting stuck looking for a perfect solution, but if you believe it's better this way, let's just close this one.

feiskyer:
> what do you think about having one file, for example releases/patch_releases.md which has a section for each supported branch.

It's a good idea to include all in one file. Much easier to find for cherry picks, as well as schedulers.

tpepper:
No need to close this one. If folks prefer one location, just amend your commit to use the one directory higher file location, and at most add one level of indention inside the file with place holder section names for 1.12 and 1.11, and maybe a table of contents at the top of the file. You've done the 1.13 content and there are owners for adding 1.12 and 1.11 content.

tpepper:
I think this is a good start.  Certainly much better than what was visible before.  We can incrementally improve and extend it.

/lgtm

PR #435, opened 2019-01-09
release lead role handbook: encourage global mindset

tpepper:
In various places our documentation on release team roles and operations
have a legacy bias towards the Pacific timezone.  This is no longer
necessary and is unhealthy.  We should strive to be a globally inclusive
community.

### DIFF
--- a/release-team/role-handbooks/release-team-lead/README.md
+++ b/release-team/role-handbooks/release-team-lead/README.md
@@ -17,7 +17,7 @@ The Release Team Lead should be an arbiter of decisions, and not the primary dec
   - You must be a member in good standing, and part of the Kubernetes org
   - You must be active in at least one Kubernetes SIG
   - You must be a member of the kubernetes-dev mailing list
-- Several release team members are in the Pacific Time Zone partially due to infrastructure access limitations at Google, therefore, the release lead must be comfortable scheduling meetings and events compatible with PT working hours
+- The Release Team is a global volunteer group, as is the Kubernetes project, so the Lead must be prepared to schedule and attend meetings that may be very late or very early in his or her time zone, in order to include as many of the team as possible.
 - Project management experience is highly desirable
 - Strong written and verbal communications skills are required
 - As the public face of the community during release time, you must do so with a very high level of professionalism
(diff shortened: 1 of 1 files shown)


release-team/role-handbooks/release-team-lead/README.md
```diff
@@ -17,7 +17,7 @@ The Release Team Lead should be an arbiter of decisions, and not the primary dec
...
-- Several release team members are in the Pacific Time Zone partially due to infrastructure access limitations at Google, therefore, the release lead must be comfortable scheduling meetings and events compatible with PT working hours
+- Historically release team members have been focused close to the Pacific Time Zone partially due to infrastructure access limitations at Google.  These limitations are most entirely legacy at this point.  Therefore, the release lead should be comfortable scheduling meetings and events compatible with global working hours and coordinating a global set of team leads and shadows.  This may mean having the traditional 10am Pacific release team meeting, but also scheduling alternate time meetings and varying the final weeks' burndown meeting time slots to best accommodate the individuals (both release team and broader SIG representation on the issues of the day) whose presence and information is needed.  Sacrifice may be necessary at times and the release lead should endeaver to spread this so as not to focus any inconvenience on specific individuals or specific geographies, while also making sure the project is correctly moving forward.
```
justaugustus: If we're going to break with tradition, let's remove the line about 10am altogether; it almost suggests that one of the meetings should still be at that time. I think it's enough to say that the RT Lead will work to find times amenable to the team.

justaugustus: `Historically,`

justaugustus: `Zone, partially`

AishSundar: s/"coordinating a global set of team leads and shadows"/"coordinating **with** a global set of team leads and shadows"

justaugustus: `endeavor`

justaugustus: `Release Team Lead`

AishSundar: Can we also include some text on the release lead taking global holidays into account when schedulng meetings and release calendar itself.

tpepper:
Comments addressed.

I've left the 10am meeting reference.  I think of this as the baseline and feel it's reasonable to keep it consistent release to release, although at the point there is an Release Team Lead from an APAC region it's bound to become necessary to shift this.  What I'm aiming to do with this commit is indicate that the Release Team Lead needs to do more than just this static base line.

release-team/role-handbooks/release-team-lead/README.md
```diff
@@ -17,7 +17,7 @@ The Release Team Lead should be an arbiter of decisions, and not the primary dec
...
+- Historically, release team members have been focused close to the Pacific Time Zone, partially due to infrastructure access limitations at Google.  These limitations are most entirely legacy at this point.  Therefore, the Release Team Lead must be comfortable scheduling meetings and events compatible with global working hours and coordinating with a global set of team leads and shadows.  This may mean having the traditional 10am Pacific release team meeting, but also scheduling alternate time meetings and varying the final weeks' burndown meeting time slots to best accommodate the individuals, both release team and broader SIG representation on the issues of the day, whose presence and information is needed.  Sacrifice may be necessary at times and the Release Team Lead should endeavor to spread this so as not to focus any inconvenience on specific individuals or specific geographies, for example considering major global holidays when scheduling, while also making sure the project is correctly moving forward.
```
jberkus: This feels very long, and I don't think the historical reference is necessary.  Counter-suggestion:

"The Release Team is a global volunteer group, as is the Kubernetes project, so the Lead must be prepared to schedule and attend meetings that may be very late or very early in his or her time zone, in order to include as many of the team as possible."

If we want to offer further advice on what this looks like, I suggest putting it in some other section of the guide than the Requirements.

tpepper: I like it.  I've used this sentence and instead moved the larger paragraph down into the time commitment section.  It's important still for the lead to be thinking about the relative commitments the broader team gets pulled into.

jberkus:
That looks really good.

/lgtm

Issue #269, opened 2018-08-31
Expand on slush definition


pbarker:
In the 8/30 sig-arch meeting the question was raised on whether api adds could be merged in slush. The consensus was a firm no. This isn't reflected in the current docs and we should consider explicitly calling out types of changes that are restricted.

justaugustus:
This is open for anyone who wants to write a quick blurb. @jdumars?
/help

spiffxp:
/milestone v1.14

It's not clear to me whether code slush has served its intended purpose.  From what I can tell, it is meant to signal that we aren't accepting any major refactors, but is intended to allow for a higher velocity of code changes than code freeze.

Does it simplify things if we just get rid of it entirely?  If we keep it around, can we more tightly quantify what is and is not allowed in a way that we can verify the effectiveness of this added process?

nikopen:
+1 on getting rid of code slush. In the current handbook iteration of bug-triage there's 'early release', 'pre-code-slush', 'code-slush', 'code-freeze' and 'burndown' - I'm in favor of fewer stages with clearer definitions while I'm cleaning up the handbook (https://github.com/kubernetes/sig-release/pull/430).

In its stead, there can be clearer timelines and boundaries on code freeze - freeze could happen a bit earlier, with a clearer signal, i.e. 'major features should be ideally finished around the first half of code freeze timeslot'.

spiffxp:
Discussed at release team this morning.  The experience of previous release team members was that code slush was a useful "code freeze is coming" deadline.  For example, some SIGs would wait until code slush to decide whether or not a feature was going to land.  Most felt a blast across kubernetes-dev@/twitter/slack would be sufficient for this.

The actual mechanism of requiring additional labels and milestones seemed maybe not worth the hassle.  Code slush as a concrete phase maybe doesn't need to exist.

PR #442, opened 2019-01-12
Remove code slush

spiffxp:
We decided a clear warning to brace for code freeze was sufficient,
and a whole extra delineated phase wasn't worth the effort. We'll
still remind folks not to do mean things like land massive changes
or refactors N days before code freeze.

ref: https://github.com/kubernetes/sig-release/issues/269


spiffxp:
/hold
for comment (also, ref: https://github.com/kubernetes/community/pull/3090)

spiffxp:
I vastly underestimated how many times the word "slush" is mentioned in playbooks.  The following playbooks need an update:
- test-infra
- bug-triage
- release-notes
- branch-manager
- release-team-lead
- ci-signal
- enhancements

I would rather hold this to take an initial pass

tpepper:
I'm also in favor of this, but definitely would prefer removal across the board so will await the amended commit / PR.

Issue #192, opened 2018-06-19
process docu: release team Docs Lead


tpepper:
The release team's Docs Lead role is only minimally documented in two bullets of text in:
https://github.com/kubernetes/sig-release/blob/master/README.md#docs-lead
and there are old docs in:
https://github.com/kubernetes/sig-release/tree/master/release-process-documentation/documentation-guides

A lot has changed lately around docs generally and also specific to release, including some really good observations and improvements around interacting with the release team's Features Lead and SIG leads.

It would be good to have old docs merged into an up-to-date doc for the role at:
https://github.com/kubernetes/sig-release/tree/master/release-process-documentation/release-team-guides

zacharysarah:
@tpepper Agreed, an updated guide would be good. I'll echo my suggestion in #139 that @mistyhacks (and maybe @zparnold as he enters the docs release lead role for 1.12) continue any updates.

mistyhacks:
Yeah I'll work on this after the smoke clears from 1.11. 👼

justaugustus:
@zparnold -- Are we all set with the Docs Handbook?

idealhack:
cc @jimangel docs lead of 1.14

jimangel:
@msohn I'll fix the link today.

@justaugustus (3 months later...) we have a couple handbooks out there but none are considered "complete." One of my goals this cycle is to correct that along with docs for shadow responsibilities.

Thanks for the CC @idealhack!

jimangel:
/unassign @zparnold
/assign @jimangel

cc @zacharysarah

PR #443, opened 2019-01-13
temp fix docs link

jimangel:
Updating the broken link for docs as a temp fix. Post 1.14 release, these docs will be updated for the 1.15 folks. I also plan on modifying the directory structure to match the other handbooks (README.md).

cc @justaugustus 

Ref: https://github.com/kubernetes/sig-release/issues/192


jimangel:
I'm good with that. Updated. I still kept in some of the extra docs which will be merged / phased out as I clean it up. Thanks!

PR #444, opened 2019-01-14
Update TODO's

spiffxp:
Went ahead and copy-pasted the 1.13 docs to 1.14 docs and made bitly links

Added info about kubernetes-release-team@


spiffxp:
/hold
I'd like to get an OWNERS file in here too

PR #445, opened 2019-01-14
Update OWNERS for release-team

spiffxp:
This mildly conflates two concerns:
- I think the OWNERS of the release-team subproject should be distinct from the SIG Release Chairs
- I think the current and former release-team leads should have /approve access to release-team playbooks

I'm opting to addres the second concern here. We can replace aliases with specific people once we've had a chance to discuss at SIG Release

### DIFF
--- a/release-team/OWNERS
+++ b/release-team/OWNERS
@@ -1,11 +1,5 @@
-reviewers:
+# See the OWNERS docs at https://go.k8s.io/owners
+
+approvers:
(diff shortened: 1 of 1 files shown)


tpepper:
Looks reasonable to me, good cleanup.

BenTheElder:
sorry about that, there's a race on /hold-ing that is rarely won :-(

PR #446, opened 2019-01-14
Revert "Update OWNERS for release-team"

calebamiles:
Reverts kubernetes/sig-release#445 as @justaugustus had concerns


justaugustus:
We worked it out on the channel. I mistook this for a change to `releases/release-1.14/OWNERS`, which didn't include the current Release Team as reviewers.

...but it's not, so I'm fine with it. :)
/close

PR #462, opened 2019-01-22
Update release_team.md

claurence:
updating release team with the enhancement shadows


claurence:
@spiffxp @justaugustus

Issue #458, opened 2019-01-21
Document kubernetes-release-team Google Groups membership criteria


spiffxp:
/kind documentation
/milestone v1.14
/sig release

My swag before I open a PR:

Who gets to be a member?
- Current release team members
- Current patch release team members

Who gets to be a manager?
- Release team lead

Who gets to be an owner?
- SIG Release leads

spiffxp:
Should shadows be in or out?

justaugustus:
+1 to membership + ACLs as you have them, @spiffxp. Pretty sure that's what I did on the last sweep.
Yes, we should add shadows, though we need to consider the whole disappearing shadow thing and prune often.

justaugustus:
Open for an RT member who wants to take a crack at this.
/help

jeefy:
/assign @lachie83 

@justaugustus Lachie'll take this on unless you really want to :)

alejandrox1:
ping @lachie83 do you have any updates?
/milestone v1.17

guineveresaenger:
Now that I know we're talking about mailing list memberships:

- All Role Leads and Shadows for RT
- SIG release chairs
- Emeritus Advisor

Does Release Management have a separate group? if not, everyone who is on that team too.

RT Lead should update this list at the start of the cycle

Chairs are responsible for making RT Leads managers

justaugustus:
/unassign
/assign @guineveresaenger

> Does Release Management have a separate group? if not, everyone who is on that team too.

We've got our own [lists](https://github.com/kubernetes/sig-release/blob/master/release-managers.md#contact), but yes, current [Branch Managers](https://github.com/kubernetes/sig-release/blob/master/release-managers.md#branch-managers) should be on this list as well to stay abreast on Release Team communications.

> RT Lead should update this list at the start of the cycle

Timing is important here. Some things I've noticed:
- It's easier to prune the list once the next team (including shadows) is locked in (since there may be people returning)
- It's good to figure out any final Release Team communications soon (swag, experience surveys, etc.)
- It's bad to send comms for the next release w/o pruning the prior team (has led to a few moments of confusion/awkwardness for shadows that may not be a part of the next cycle)

> Chairs are responsible for making RT Leads managers

+1

alejandrox1:
Will work on this this week

Issue #440, opened 2019-01-12
Consider using kubernetes-release-team@ instead of kubernetes-milestone-burndown@


spiffxp:
We currently use a [kubernetes-milestone-burndown@] google group for the meeting invite for release team meetings, and in general as a google group for ACLs related to the release team.  But it's really got so many more people than just the release team on it.

It turns out there is a [kubernetes-release-team@] google group that didn't really get documented as part of the 1.12 cycle.  I noticed it when I was pruning acls for https://bit.ly/k8s114-enhancements. I'd like to pick it back up for the 1.14 release cycle.  Release team leads and shadows should get added here.  cc @justaugustus 

Need to decide/document how this is used.

[kubernetes-release-team@]: https://groups.google.com/forum/#!forum/kubernetes-release-team
[kubernetes-milestone-burndown@]: https://groups.google.com/forum/#!forum/kubernetes-milestone-burndown

/milestone v1.14
/kind cleanup
/sig release

spiffxp:
Step 1: I'd like to purge membership of kubernetes-release-team@ to match that of the v1.14 release-team + release-team subproject owners (let's call that sig-release leads for now) + patch-release team.  Long term I feel like the patch-release team should be split into its own entity

justaugustus:
As the person who created the group, I'm definitely in favor of this :)
/assign

justaugustus:
The Kubernetes Release Team list membership has been updated.
After some discussion on Slack, it seems that the milestone-burndown serves no purpose that couldn't just as well be served by the top-level sig-release list.

That said, I'd like to get rid of it and @spiffxp is in agreement.

I'll let the other SIG Release Chairs chime in, on the off-chance I'm missing some context and then I'll put the request official on k-dev, k-milestone-burndown, and k-sig-release.

Lazy consensus timebox EOD PT Monday, 1/28.
/assign @calebamiles @tpepper

tpepper:
I'm good with this.

spiffxp:
Here's my proposal for use of the lists as it relates to calendars and docs:
- ensure sig chairs have owner privileges to the kubernetes-release-team@ and kubernetes-milestone-burndown@ lists
- release doc ACL's should be as follows
  - release team contact info should be readable only by kubernetes-release-team@ and kubernetes-sig-release-leads@
  - all other release related docs should be commentable by everyone
  - all other release related docs should be writable by kubernetes-release-team@ and kubernetes-sig-release-leads@
[...]

PR #480, opened 2019-01-30
Create exception.yml for 1.14 release 

claurence:
creating an exception file for the 1.14 release enhancements


claurence:
@spiffxp as an FYI - making an exception.yml for 1.14 now that we've past enhancements freeze

releases/release-1.14/exception.yaml
```diff
@@ -0,0 +1,12 @@
+# Exceptions to Code Freeze requested in 1.14
+# Google Group: https://groups.google.com/forum/#!topic/kubernetes-milestone-burndown
```
spiffxp: I think we should use kubernetes-release-team instead of -burndown for these requests

spiffxp:
ref https://github.com/kubernetes/sig-release/issues/458 and https://github.com/kubernetes/sig-release/issues/440 for the release-team@ issue

releases/release-1.14/exception.yaml
tpepper: I've toggled kubernetes-release-team to "anyone can post".

justaugustus:
We can clean this up when we turn down k-m-b@
/lgtm
/approve
/hold cancel

PR #476, opened 2019-01-29
[wip] clarify release-1.14 docs around enhancements freeze

spiffxp:
As I'm working on this, I realize that the "exceptions" doc as written
does _not_ really apply to enhancements that didn't make it in by enhancements
freeze. I can't decide whether we should not link to it, and make up a process
on the fly... or heavily reword it

ref: https://github.com/kubernetes/sig-release/blob/master/releases/EXCEPTIONS.md

PR #482, opened 2019-01-30
Release 1.14 readme week4 updates

liggitt:
Builds on https://github.com/kubernetes/sig-release/pull/476 to add https://github.com/kubernetes/sig-release/pull/476/files#r252319391


liggitt:
cc @claurence @spiffxp

PR #507, opened 2019-02-19
remove 'none' line

nikopen:
seems that when `/release-note none` is applied to a kind/feature PR, it is fetched as 'none' in the release notes.

https://github.com/kubernetes/kubernetes/pull/74182#issuecomment-464554714


marpaia:
/lgtm
/approve

We will need to update the regex in the `release-notes` tool to ensure that this doesn't keep happening.

PR #526, opened 2019-02-26
Add role-handbooks README

spiffxp:
Presently github is displaying the docs/README.md in the UI whenever
users visit the role-handbooks directory. I am hoping an actual
README.md in role-handbooks' root will override that.

/kind documentation


spiffxp:
https://github.com/kubernetes/sig-release/blob/master/release-team/role-handbooks - it worked!

PR #561, opened 2019-03-22
Add communications shadows for 1.15 release.

onlydole:
Add the following members as communications shadows for the 1.15 release:

* Taylor Dolezal
* [Open Slot]


castrojo:
Ok so someone brought up a good point with me on slack.  With release trying to keep the amount of shadows down to 2, we should think about the future and perhaps keep one open. 

@nwoods3 and @kbarnard10 since you have to be here for CNCF PR _anyway_  how about we do taylor and an open slot with the goal of trying to get a new person in?

PR #569, opened 2019-03-27
Track containers not rebased to distroless

yuwenma:
As part of the effort on sig-release KEP "[Rebase k8s images to distroless](https://github.com/kubernetes/enhancements/pull/900/files#)", we allow certain kinds of images rebasing to k8s.gcr.io/debian-base instead of distroless for now. This list is used to track containers not rebased to distroless in v1.15.

### DIFF
--- a/release-engineering/baseimage-exception-list.md
+++ b/release-engineering/baseimage-exception-list.md
@@ -1,0 +1,5 @@
+# Base Image Exception List
+
+| Component Name        |           Base Image         |        Reasons    |
+| --------------------- | :---------------------------:|:------------------:|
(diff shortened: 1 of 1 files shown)


tallclair:
Thanks for adding this. I missed it when it went in, but can you add a small heading that explains the context for this list? Maybe an brief explanation and a link to the KEP

yuwenma:
> Thanks for adding this. I missed it when it went in, but can you add a small heading that explains the context for this list? Maybe an brief explanation and a link to the KEP

Gotcha, will update the list and add more descriptions.

Issue #454, opened 2019-01-20
Release Team survey improvements


justaugustus:
Now that we have a survey for Release Team volunteers, I'd like to gather suggestions on how we can:
- Improve the survey
- Easily massage the survey data in future

@BenTheElder also brought up a great point: while we won't be able accept everyone on the Release Team, this seems like it could be an additional resource for seeding volunteers for other places in the project.

I'm happy to have anyone who wants to carry this issue forward pick it up.
/help

/sig release contributor-experience

cc: @calebamiles @tpepper @spiffxp @BenTheElder @marpaia @parispittman @idvoretskyi @castrojo  @nikhita

idvoretskyi:
> * Improve the survey

@justaugustus can you please share the currently used version?

jberkus:
Hey, I'm taking this over, and I have no access to the original survey.  Will pester people on Slack.

jimangel:
I solicited feedback throughout the v1.14 release cycle and summarized what I found in a [google doc ](https://docs.google.com/document/d/1tBRTh9YdGQQgy_8nUOnkIhadhssny8XkYxuB58mytaw/edit)today. I then sent that doc out to every v1.14 release lead and encouraged them to share it with their shadows. This was the net takeaway:

The survey should limit the candidate from selecting <2 roles (default for 1.14).

Questions the community would like to add:
1. How many hours a week do you think you'll need to fulfill these duties?
1. Which part about the role(s) interest you the most?
1. Which part about the role(s) interest you least?
1. What are you looking to gain from shadowing the release team?
1. What are your future goals for contributing within the Kubernetes community?
1. Communication preference (multiple choice of Github / Slack / email)?

Definition of the shadow selection process as lead:

- Review
  - you will review the survey results
  - you will review candidates background
    - Github / LinkedIn / Other
  - you will conduct 1:1 interviews with candidates they are interested in
    - Interviews will be over Slack / email / Zoom (your preference) 
- Selection (after reviewing [the guide](https://github.com/kubernetes/sig-release/blob/master/release-team/release-team-selection.md))
  - There is no recommended number of shadows to select.
    - Use your best judgement of what you feel comfortable with mentoring
    - When in doubt, use the previous release as a guideline
  - Reach out to your selected candidate(s)
- Rejection
  - Communicate (email or Slack) to anyone not selected using a defined template.
[...]

PR #576, opened 2019-04-10
Updating shadow selection timeline

jimangel:
Clarifies when future release team is selected, when survey is published, when it's closed, and when release team should be 100% locked in.

Note: This PR references "Emeritus Adviser" which may need to be updated in the future with who is responsible for releasing the survey (anyone _except_ the release team lead)

Ref: https://github.com/kubernetes/sig-release/issues/454

We might also want to consider branding it as a "questionnaire" or a "survey" verses using the terms interchangeably.

/cc @jberkus @justaugustus


jberkus:
"submission form"?
"participation request"?
"webformwithawholebunchaquestions"?

PR #608, opened 2019-05-01
Docs handbook updates

jimangel:
Closes: https://github.com/kubernetes/sig-release/issues/579 and double closes: https://github.com/kubernetes/sig-release/issues/192 haha

Edit: Link to doc file for quick review: https://github.com/jimangel/sig-release/blob/docs-handbook-updates/release-team/role-handbooks/docs/README.md

/cc @justaugustus @cody-clark

PR #613, opened 2019-05-06
finalizing docs handbook

jimangel:
Addresses remaining nits from https://github.com/kubernetes/sig-release/pull/608

### DIFF
--- a/release-team/role-handbooks/docs/README.md
+++ b/release-team/role-handbooks/docs/README.md
@@ -2,7 +2,7 @@
 
 ## Overview
 
-This document covers the responsibilites, time commitments, and timeline for Docs Leads sheparding docs releases for Kubernetes. Docs Lead Shadows should also read through this document and understand the launch processes so they can do it in the future.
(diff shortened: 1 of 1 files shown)


release-team/role-handbooks/docs/README.md
```diff
@@ -350,6 +350,10 @@ Create the PRs needed to roll the docs to the new version on the dev branch
 
 1. Deprecate the oldest API link in the reference docs, e.g: https://github.com/kubernetes/website/pull/13467
 
+1. Update the index for the API reference docs, e.g: https://github.com/kubernetes/website/pull/14139
+
```
kbhawkey: Do you think more details are warranted:
- Deprecate the oldest API link and add a new link for this release version, ....
    There should be five links to the latest five releases.
- Update the API reference version, in the API reference index page, to match the release version, ...

kbhawkey: That's fine!
Is there a link to this handbook from a page in /docs/contribute? Good for docs people to stay involved.

MAKOSCAFEE: > I think the changes are minor enough that an example would suffice. However, I will defer to @MAKOSCAFEE for the final say. In the meantime, I think we can merge this "as is" and Barney can include it if he chooses with any of his revisions post-release.

Sure this is fine. Will add more examples on my revisions.

PR #620, opened 2019-05-09
patch release handbook: add workflow reference example

tpepper:
The order of operations in the release workflow is a mix of complicated,
error-prone command invocations, and manual synchronization check
points.  A summary of these is needed to make the required actions more
clear.

Also adds a table of contents as intradoc links makes eaiser navigating to the
appropriate info a different points in a process that spans days or even
weeks.

Signed-off-by: Tim Pepper <tpepper@vmware.com>

PR #621, opened 2019-05-09
[WIP] makefile wrapping concept

tpepper:
The commit looks fugly, but basically is adding a third column to the table introduced in https://github.com/kubernetes/sig-release/pull/620

For a concrete discussion on tools improvement, I want to show what I'm imagining it might look like to establish a wrapping of the detailed commands which do build and publish.

My goal is to allow parallel back end implementations and proof-of-concept comparisons.  If the top level flow is abstracted and callable by automation, without those tools getting coupled to the internal implementation, we are in a more flexible and testable place and enable more to be done through automation.


tpepper:
/hold

here for discussion

justaugustus:
/close
to revisit in handbook rewrites

PR #662, opened 2019-06-12
patch release schedule: June 2019 update

tpepper:
1.13 and 1.14 next patch releases are likely later this month.  1.15
is getting close, with a first patch release likely early next
month.

Hannes Hoerl is broadening the patch release team out a notch.


tpepper:
> That said, all listed Patch Release Managers support all supported Kubernetes versions, so that column is unnecessary.

removed

Issue #668, opened 2019-06-14
Adjusting PR boilerplate and labels to improve Release Notes UX


onyiny-ang:
There was some discussion during this release cycle and in [slack](https://kubernetes.slack.com/archives/C2C40FMNF/p1560535286311000 ) about collecting more information from each PR in order to improve the user experience for release notes and ease curation for the release notes team.
This boils down to 3 main improvements:
1. Collecting `KEP` numbers associated with any PRs that work towards implementing a KEP.
2. Collecting the `Primary-sig` or sponsoring SIG associated with each PR to eliminate guess work for the release notes team.
3. Labeling the PR as a deprecation if it is a deprecation to a dependency.

I would like to propose a subform--or in any case, more detailed release note section of the k/k PR boilerplate that ensures a contributor fills out the `Primary SIG` section and has a space to associate a `KEP` # if it is appropriate. A label to indicate deprecations would be great.

onyiny-ang:
/cc @tpepper @justaugustus @jeefy @saschagrunert @lachie83 @jberkus 
/sig release

Ideally we would like to have some (or all) of these things available starting in the 1.16 release

saschagrunert:
I think the only related PR is https://github.com/kubernetes/kubernetes/pull/81278, which we either have to finish or postpone to the next release.

saschagrunert:
Please take a look at the two linked PRs, we decided to split up the work a bit more.

guineveresaenger:
We are waiting on some automation and decisions around proposed labels here.

This won't make the 1.17 milestone.

/milestone v1.18

justaugustus:
I think we can close the loop on this before 1.17 is out.
/milestone v1.17

saschagrunert:
I propose that we close this issue now. If we're going to sort the notes by kind then we would not need a primary SIG at all, WDYT?

justaugustus:
This needs to stay open until we have a community decision on kind vs SIG.
See here: https://github.com/kubernetes/release/pull/923#issuecomment-558880526

I'll send a note out next week since US folks are on holiday now.
/remove-help
/unassign
/area release-eng

saschagrunert:
@justaugustus I think we can close this issue, what do you think?

alejandrox1:
Just confirmed with @justaugustus that it is cool to close this one
/close

Issue #667, opened 2019-06-14
Add links to relevant docs in Release Notes


jeefy:
Per https://kubernetes.slack.com/archives/C2C40FMNF/p1560536323319700

It'd be awesome if, in the release notes, we link to new and/or relevant documentation on the K8s website. I think this is entirely reasonable and something that should be looked into in the next release or so. 

With the website it would be relatively easy and painless to add a new (optional) field in the JSON. That way it could be captured by either a field in the PR template OR a separate PR to update the JSON with the doc(s) link.

saschagrunert:
How about adding an additional section to the PR template, which could be parsed by the release-notes tool afterwards:

```
\   ```documentation
\    - Link
\    - Link
\    - Link
\   ```
```

pswica:
@saschagrunert 
I think it's a great idea. 

I think we would need to parse the links to make sure that they are on the Kubernetes website. Other things to keep in mind are links changing on the website overtime. A final thing thing to address would be irrelevant links. What if people added a link to the main kubernetes website for their PR. Would that really be relevant if it appeared dispersed over 20 PRs or so?

saschagrunert:
> What if people added a link to the main kubernetes website for their PR. Would that really be relevant if it appeared dispersed over 20 PRs or so?

Yeah I guess we have to review the PRs with keeping in mind that the documentation links will appear in the release-notes and need to  add valuable benefits to the users. Nevertheless we can parse all URLs and classify them by the location. So we would be able to separate KEP links or filter only for KEPs inside the website.

saschagrunert:
@jeefy do we want to close this or is there anything else which needs to be done? :)

jeefy:
🌮 

/close

PR #673, opened 2019-06-16
Release Notes Role handbook update

onyiny-ang:
Update to Release Notes Role handbook pre-1.16 release kick off.


release-team/role-handbooks/release-notes/README.md
```diff
@@ -66,6 +69,7 @@ beginning of code freeze and the release.
 ## Tools
 
 - [Release notes tool](https://github.com/kubernetes/release/tree/master/cmd/release-notes)
+- [Release notes website](https://relnotes.k8s.io)
 - [Hackmd](https://hackmd.io/)
 - [LWKD](http://lwkd.info) *(consider contributing to LWKD as part of your role)*
```
saschagrunert: Do we want to add the TODOs for 1.15 (or convert the ones from 1.14) as well? :point_down:

onyiny-ang: Yes. I think two of the todos are definitely #668 and #667 and then we should work out the others before this goes live

saschagrunert: Sounds good, these two should be definitely addressed in the next cycle.

saschagrunert:
LGTM, do we need approval from anyone else or can I simply drop my ltm here? :innocent:

PR #676, opened 2019-06-17
Add 1.16 RT lead-shadows

lachie83:
cc @guineveresaenger @nikopen @jeefy


justaugustus:
Enthusiastic +1 for these RT Lead Shadows! :)
/lgtm
/approve
/pony party
/capybara

PR #716, opened 2019-07-08
Add Stephen Augustus as second EA because that's materially true.

jberkus:
Stephen has handled the whole Shadow selection, so he should be listed here.

/assign @lachie83


justaugustus:
/lgtm
/approve

Thanks @jberkus!!

PR #722, opened 2019-07-16
Created file release-engineering/artifacts.md

bartsmykla:
Document was created after a long discussions with
@justaugustus and @tpepper and I think it will be helpful
for everybody in the community to have full knowledge
about all artifacts which are being built during the release
process.

If anyone is interested in looking into raw files which
I used to prepare this document (results of `ls -lah`
commands, listing of all files inside GCS bucket after
GCP finished building artifacts etc.) you can check
https://github.com/bartsmykla/k8s-artifacts

File contains all artifacts which are being built during
the release process. There are docker images, binaries,
tar archives and RPMs/Debs

Signed-off-by: Bart Smykla <bsmykla@vmware.com>

### DIFF
--- a/release-engineering/artifacts.md
+++ b/release-engineering/artifacts.md
@@ -1,0 +1,698 @@
+## Docker Images
+
+|                          	| 386 	| amd64 	| arm 	| arm64 	| ppc64le 	| s390x 	|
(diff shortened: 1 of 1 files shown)


justaugustus:
(I mentioned this via DM, but echoing here as well...)

Thank you immensely for working on this, @bartsmykla!
I was doing a similar investigation and it's nice to have this all captured in one place.

/lgtm
/approve

PR #727, opened 2019-07-17
Add initial release notes draft for 1.16

kcmartin:
The initial draft of release notes for 1.16

start sha = 02dc3d713dd7f945a8b6f7ef3e008f3d29c2d549
end sha =  704790e0412842e782904f1c2e7123112a9aab2c

### DIFF
--- a/releases/release-1.16/release-notes-draft.md
+++ b/releases/release-1.16/release-notes-draft.md
@@ -1,0 +1,274 @@
+## Action Required
+
+- Migrate scheduler to use v1beta1 Event API. any tool targeting scheduler events needs to use v1beta1 Event API ([#78447](https://github.com/kubernetes/kubernetes/pull/78447), [@yastij](https://github.com/yastij)) Courtesy of SIG API Machinery, SIG Apps, SIG Auth, SIG Scheduling, and SIG Testing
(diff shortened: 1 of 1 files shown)


justaugustus:
/lgtm
/approve

@saschagrunert @jeefy @onyiny-ang -- can we investigate why `NONE` got picked up as a release note [here](https://github.com/kubernetes/sig-release/pull/727/files#diff-57eb86b9b211aeea24ad87b5c47a6306R59)?

onyiny-ang:
I will double check the release notes tool later but afaik, it doesn't actually check the content of a release note so if the contributor writes "NONE" in the section for release notes (as you can see was the case for this PR https://github.com/kubernetes/kubernetes/pull/79637), the release notes tool picks it up.
We could adjust it so that it filters out release notes with content that matches a select set of commonly used "none"-like terms :)

paulbouwer:
@onyiny-ang @justaugustus @saschagrunert @jeefy @kcmartin It looks like there is a content filter on variations of NONE in the `release-notes` tool at [/pkg/notes/notes.go#L429-L445](https://github.com/kubernetes/release/blob/master/pkg/notes/notes.go#L429-L445). 

There are 2 PRs that slipped through during this rendering of the release notes with "none" in the release notes block:
- [PR #79637](https://github.com/kubernetes/kubernetes/pull/79637)
- [PR #78559](https://github.com/kubernetes/kubernetes/pull/78559)

I'll have a look later and see if we can tighten up the rules.

saschagrunert:
Okay good, thanks for taking care of the issue @onyiny-ang @paulbouwer

PR #770, opened 2019-08-22
patch release schedule update

tpepper:
Note the next update versions, but leave dates TBD for now (sometime
mid-September likely).


tpepper:
@kubernetes/patch-release-team if we're ready to declare a target date in September, I can amend this PR to include that.  Otherwise the version strings are now out of date.

PR #786, opened 2019-09-12
Release notes handbook update

saschagrunert:
Let's collect all the things we want to carry over to the next release.

/cc @onyiny-ang @paulbouwer @kcmartin @cartyc


saschagrunert:
I don’t see more things which needs to be changed for now. Let’s continue with the overall release notes automation in the scope release-engineering.

PR #811, opened 2019-09-25
Release notes Role: Fix spelling

mrueg:
Found those while reading more about the roles.


cartyc:
/approve

Thanks for catching those!

PR #825, opened 2019-10-08
Update Release Notes Handbook-Patch Release notes for website

kcmartin:
Procedure for adding Patch Release notes to the Release Notes website


release-team/role-handbooks/release-notes/patch-notes.md
```diff
@@ -0,0 +1,61 @@
+### Generating Patch Release Notes for the Release Notes website
+
+Note: There is likely an opportunity to automate this process :)
```
saschagrunert: I think we're already on the way. :grin: Do we want to mention the now easier approach to generate them, like this:

```bash
release-notes
    --branch=release-1.15
    --github-token X
    --repo-path /tmp/k8s
    --release-version 1.15.3
    --format json
    --output release-notes-1.15.3.json
    --start-rev v1.15.2
    --end-rev v1.15.3
```

This way we would not have to manually choose the revisions from the website. My plan was to implement a workflow based approach for the patch notes to reduce the required information even further, like described here: https://github.com/kubernetes/release/pull/880

cartyc: I think having the new option in there would be great :)

cartyc:
Looks great @kcmartin ! I think adding the change @saschagrunert made would be an excellent addition. Otherwise the doc looks fantastic :)

kcmartin:
Thanks @cartyc and @saschagrunert !
See update :)

PR #835, opened 2019-10-21
feat(comms): add note about Slack channel for communications team

rawkode:


justaugustus:
(@onlydole -- as a heads-up, `/lgtm` needs to be on a line by itself.)

PR #845, opened 2019-11-01
add machine requirement to run branchff

cpanato:
When running `branchff` in my local vagrant machine it took a long time to run and failed due out of memory.

checking the config it was only 1Gb and 1 vcpu to that VM.
- spin a node in DO with 4vcpu/8gb and it ran ok and did another test with 6vpcu/16Gb which this last one was much better.

adding a note in the Branchff for the machine requirement to not loose time as me :)

/area release-team
/sig release
/cc @kubernetes/release-team
/cc @justaugustus

### DIFF
--- a/release-engineering/role-handbooks/branch-manager.md
+++ b/release-engineering/role-handbooks/branch-manager.md
@@ -552,6 +552,10 @@ Prior to running `./branchff`, you will need:
 - permission to run under `root` privileges
 - membership in the [Release Managers GitHub Group](https://github.com/orgs/kubernetes/teams/release-managers)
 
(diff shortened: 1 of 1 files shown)


justaugustus:
/hold
I started a [discussion in Slack](https://kubernetes.slack.com/archives/C1TU9EB9S/p1572627769216100), as I'd like us to defer to some canonical reference point for system requirements for Kubernetes developers, if one exists.

justaugustus:
branchff no longer updates the OpenAPI spec, so this isn't required anymore.

ref: https://github.com/kubernetes/release/pull/921
/close

PR #848, opened 2019-11-05
Fix prow config whitelist links

saschagrunert:
Just a small nit which came across my eyes.


justaugustus:
Good catch. That config location recently changed.
/lgtm
/approve

PR #858, opened 2019-11-14
Update  CI signal handbook with the new github team

alenkacz:


alenkacz:
cc. @justaugustus

alenkacz:
Related to https://github.com/kubernetes/org/pull/1404

release-team/role-handbooks/ci-signal/README.md
```diff
@@ -77,6 +83,7 @@ Here are some good early deliverables from the CI Signal lead between start of t
     - For failing tests: `[Failing Test]: testgrid-tab-name (prow-job-name)`
     - For flaky tests: `[Flaky Test]: testgrid-tab-name (prow-job-name)`
   - Make sure all open issues have a `priority/` label (see: [Priority Labels](#priority-labels)) and one of either the `kind/flake` or `kind/failing-test` label
+  - cc. @ci-signal github team on the issue to let rest of the team aware of it
```
hasheddan: Very minor suggestion but below we say "**CC** the release manager..." and here we say "**cc.** @ci-signal github team...". I don't think it is a huge thing, but just trying to think if someone was very new to k8s / github interactions, it could feel as though these were two different types of actions. What would you think of making them either both *cc.* or both *CC*?

alenkacz: Absolutely agree. But you actually got me thinking - who is the release manager? Is that the current release lead? 🤔 or is that this group of people? https://github.com/kubernetes/sig-release/blob/master/release-managers.md what would be the case where you want to ping them? 🤔 almost wondering if we can drop it or change it in some way

hasheddan: Great point! I feel like it should maybe be release team lead?

alenkacz: Okay, updated. Thanks

release-team/role-handbooks/ci-signal/README.md
```diff
@@ -81,10 +87,11 @@ Here are some good early deliverables from the CI Signal lead between start of t
   - Assign the issue to appropriate SIG using /sig label
   - If you are aware of the individual associated with the enhancement area or issue, @mention of individual(s) and SIG leads tends to result in faster turn around
   - Add @kubernetes/sig-foo-test-failures to draw SIG-foo’s attention to the issue
-  - CC the release manager and bug triage lead if the issue needs attention
+  - CC @ci-signal github team on the issue to let rest of the team know about it, you might also CC the release lead and bug triage lead if the issue needs extra attention
     immediately
```
hasheddan: Looks like this may have snuck down to a new line :)

alenkacz: you have a good eye :)

alenkacz: It actually originally was on newline, I don't think that was intentional :)

PR #859, opened 2019-11-15
Move daily burndown back a week due to KubeCon

guineveresaenger:
We don't need to attempt daily burndowns during KubeCon


justaugustus:
Agreed!
/lgtm
/approve

PR #883, opened 2019-12-05
patch release schedule update for January 2020

tpepper:
With the release of 1.17, 1.14 falls out of support in Dec. 2019.
Typically a .1 release (ie: 1.17.1) might come two weeks after the .0
release, but for 1.17 that hits the end of year holidays.  For that
reason we're targetting 1.17.1 along with the other patch releases in
January 2020.


justaugustus:
This has multiple Patch Release Team approvals, so releasing the hold.
/hold cancel

PR #914, opened 2019-12-12
Add intro team meeting to Week 3

karenhchu:
Signed-off-by: Karen Chu <karen.chu@microsoft.com>

### DIFF
--- a/release-team/role-handbooks/communications/README.md
+++ b/release-team/role-handbooks/communications/README.md
@@ -52,6 +52,7 @@ There is a channel on the Kubernetes Slack workspace, `release-comms`, which is 
 
 - Begin analyzing the release notes draft
 - Work with the enhancements lead to understand what might be delivered
+- Establish initial meeting with team to introduce everyone, collective tasks, and upcoming timelines
 
 ### Week 4
 


release-team/role-handbooks/communications/README.md
```diff
@@ -52,6 +52,7 @@ There is a channel on the Kubernetes Slack workspace, `release-comms`, which is
 
 - Begin analyzing the release notes draft
 - Work with the enhancements lead to understand what might be delivered
+- Establish initial meeting with team to introduce everyone, collective tasks, and upcoming timelines
```
cpanato: Why in the week 3? And not earlier.

guineveresaenger:
/lgtm

Thank you Karen! 🎉

release-team/role-handbooks/communications/README.md
guineveresaenger: because shadows tend to be finalized in Week 2, depending

I take your point, though - earlier could be better!

cpanato: Ah! Ok makes sense then. Thanks for the clarification @guineveresaenger

PR #962, opened 2020-01-18
branch management: don't build at an implicit head

tpepper:
We're finding that anago is interesting with respect to how
it picks where to tag and build a tree.  Builds with --build-at-head
are going to be disabled and are bad practice because they are implicit
and opaque when build actions are logged, eg:

 ./gcbmgr stage release-1.17 --buildversion=v1.17.1-beta.0.42+d224476cd0730b

versus

 ./gcbmgr stage release-1.17 --build-at-head

The prior if re-run will (or...we can hope) be repeatable.  The latter
definitely depends on the state of the internet and repeated runs
definitely make for different outcomes.  That makes analysis and debugging
of failures harder than it needs to be.


justaugustus:
Great follow up to this week, @tpepper!
/lgtm
/approve

PR #1072, opened 2020-04-27
adjust k8s 1.19 release timeline in comms handbook

mkorbi:
This PR updates the temporarily valid timeline for the Kubernetes 1.19 release and the corresponding communications timeline.

### DIFF
--- a/release-team/role-handbooks/communications/README.md
+++ b/release-team/role-handbooks/communications/README.md
@@ -47,81 +47,183 @@ There is a channel on the Kubernetes Slack workspace, `release-comms`, which is 
 
 ## Release Milestone Activities
 
-### Week 1
+<table>
+    <tr>
+        <td><h3><b>Release Week</h3></b></td>
+        <td><h3><b>RW 1.19</h3></b></td>
+        <td><h3><b>Milestones</h3></b></td>
+        <td><h3><b>Activities</h3></b></td>
+    </tr>
+    <tr>
+        <td>1</td>
+        <td>1</td>
+        <td><b>Release Start</b></td>
+        <td>
+        <ul>
+        <li>Select your shadows
+        <li>Ensure yourself and your shadows are entered into the release contact sheet
+        <li>Ensure yourself and your shadows are on the release team meeting invites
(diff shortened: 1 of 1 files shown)


mkorbi:
Heyho @onlydole @mrbobbytables @jeremyrickard 
do you have some comments for me?

/cc @mkhaas

mkorbi:
hey @mrbobbytables can we close this?

release-team/role-handbooks/communications/README.md
```diff
@@ -47,81 +47,183 @@ There is a channel on the Kubernetes Slack workspace, `release-comms`, which is
 
 ## Release Milestone Activities
 
-### Week 1
+<table>
+    <tr>
+        <td><h3><b>Release Week</h3></b></td>
+        <td><h3><b>RW 1.19</h3></b></td>
```
jeremyrickard, line 53: I think it might be better to just drop this column and update the actual release week (and then we can do this every release). 

1.20 will probably also have a non-standard release cycle, so who ever leads comms will probably need to update this column (and rename it) for 1.20..what if we just update the calendar every release and take into consideration that future release cadence might change.  Make it a week 1 task for future releases as well.

onlydole:
@mkorbi - can you take a look at @jeremyrickard‘s comments and we can get this PR in a mergable state 👍🏼

release-team/role-handbooks/communications/README.md
mkorbi: also thought about it, but we intended to make the comms more clear and transparent. Therefore I think it would be beneficial to have the 13-week cycle as a reference, as we drove the last release very well with it. 

Just don't want to cause too many actions, that in 2021 the comms team needs to figure out at which time they have to communicate, publish and trigger.

onlydole: Do you feel good on that front, @jeremyrickard? If so, we can merge this in.

jeremyrickard: Sorry! I thought I had replied here.  2021 comms teams can always revisit.

PR #1149, opened 2020-07-15
Add graphic about code thaw and code freeze

saschagrunert:
#### What type of PR is this:

/kind documentation

This should visualize the differences between Code Freeze and Code Thaw
when putting them into the release cycle.

None

None


LappleApple:
Love all these diagrams :)

saschagrunert:
> Looks to be rendering a bit strangely for me. It looks like about a fourth of the image does not have background and the rest has white background. Since this will be rendered in a README not sure if it makes much of a difference though

Should be fixed now, that seems a bug in LucidChart.

PR #1151, opened 2020-07-15
update ref to checkout for packaging debs/rpms

listx:
This is a temporary fix; we should also move these docs to
github.com/kubernetes/release as well so that the tools and the
instructions are versioned together.

#### What type of PR is this:

/kind documentation

Fixes #

/cc @BenTheElder


BenTheElder:
thanks!

PR #1163, opened 2020-07-28
Update CI Signal Issue format to be more conducive to triage

### LINKED ISSUE #1162: CI signal handbook: adjust issue guidance to target specific SIGs
#### What would you like to be added:

Adjust the CI signal handbook guidance and templates to open one issue per SIG per set-of-failing-tests, rather than one issue per failing/flaking job.

#### Why is this needed:

The CI signal handbook gives [examples of issue templates to open for failing test jobs](https://github.com/kubernetes/sig-release/tree/master/release-team/role-handbooks/ci-signal#routing-test-issues-to-sigowner) to route issues to sig owners, but the title and instructions describe opening an issue per job.

* `[Failing Test] testgrid-tab-name (prow-job-name)`
* `[Flaky Test] testgrid-tab-name (prow-job-name)`

For jobs with tests from multiple sigs (like integration / e2e), a description mentioning one SIG's failing test can dissuade other SIGs from investigating their failures in the test.

Example: 
https://github.com/kubernetes/kubernetes/issues/92360 was opened because the integration job flaked on `cmd/kubeadm/test/cmd TestCmdInitConfig/can_load_current_component_config`, which is owned by cluster-lifecycle

That flake was resolved, but the job was then flaking on sig-scheduling failures, which kept the issue from being closed, but no one from sig-scheduling was investigating the issue.

Ideally, there would be one issue per SIG per set-of-failing-tests, with a test-grid and triage query that captures the flakes/failures the sig needs to resolve. For example, https://github.com/kubernetes/kubernetes/issues/93485 was opened, targeted at sig-scheduling, sig-filtered testgrid and triage links.

hasheddan:
#### What type of PR is this:

/kind documentation

This updates the documentation on openining issues for the CI Signal
team based on feedback from SIGs. The primary changes include
centralizing all information for opening issues to a single section, as
well as providing a decision tree that informs a team member whether
they should open one or multiple issues, and how each issue should be
named.

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>

Fixes #1162 

Examples will be added back once there are issues that reflect this new naming pattern. Note that issues are now per test instead of per job and denote the appropriate SIG in the title.

/cc @liggitt @alejandrox1

### DIFF
--- a/release-team/role-handbooks/ci-signal/README.md
+++ b/release-team/role-handbooks/ci-signal/README.md
@@ -81,20 +81,7 @@ Right after the CI signal release team is formed, CI signal lead is responsible 
 Here are some good early deliverables from the CI Signal lead between start of the release to enhancement freeze.
 - Start maintain the [CI signal project board](https://github.com/orgs/kubernetes/projects/11) and keep it up-to-date with issues tracking any test failure/flake
 - Copy over any open test issues from previous release, assign a member of the CI signal team, and have that member follow up on the issue with owners
-- Monitor [master-blocking](https://k8s-testgrid.appspot.com/sig-release-master-blocking) and [master-informing](https://testgrid.k8s.io/sig-release-master-informing) dashboards **twice a week** and ensure all failures and flakes are tracked via open issues
-  - Make sure all issues are titled either:
-    - For failing tests: `[Failing Test]: testgrid-tab-name (prow-job-name)`
-    - For flaky tests: `[Flaky Test]: testgrid-tab-name (prow-job-name)`
-  - Make sure all open issues have a `priority/` label (see: [Priority Labels](#priority-labels)) and one of either the `kind/flake` or `kind/failing-test` label
-  - Make sure the issue is assigned against the current milestone 1.x, using /milestone
-  - Assign the issue to appropriate SIG using /sig label
-  - If you are aware of the individual associated with the enhancement area or issue, @mention of individual(s) and SIG leads tends to result in faster turn around
-  - Add @kubernetes/sig-foo-test-failures to draw SIG-foo’s attention to the issue
-  - CC @ci-signal github team on the issue to let rest of the team know about it, you might also CC the release lead and bug triage lead if the issue needs extra attention immediately
-  - Assign the issue to yourself or recruit another member of the CI signal team to own the issue. The CI signal team member assigned to an issue is responsible for driving it to resolution alongside the assignee from the appropriate SIG
-  - Post the test failure in SIG’s Slack channel to get help in routing the issue to the rightful owner(s)
-  - [Sample test failure issue](https://github.com/kubernetes/kubernetes/issues/81191)
-  - Add the issue to [CI signal board](https://github.com/orgs/kubernetes/projects/11) under "New". The CI signal team lead is responsible for making sure every issue on the CI signal board is assigned to a member of the CI signal team and is being actively driven to resolution
+- Monitor [master-blocking](https://k8s-testgrid.appspot.com/sig-release-master-blocking) and [master-informing](https://testgrid.k8s.io/sig-release-master-informing) dashboards **twice a week** and ensure all failures and flakes are tracked via open issues. See [Opening Issues](#opening-issues) for how to write an effective issue.
 - Build and maintain a document of area experts / owners across SIGs for future needs e.g.: Scalability experts, upgrade test experts etc
 
 #### **_Best Practice:_**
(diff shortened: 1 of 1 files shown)


liggitt:
Thanks for the quick update. Thinking about this more, we might need a decision tree of sorts, something like:

1. If the whole job is owned by one SIG (e.g. cluster-lifecycle owns capg-conformance-v1alpha3-k8s-master / https://github.com/kubernetes/kubernetes/issues/93309 or node owns pull-kubernetes-node-e2e / https://github.com/kubernetes/kubernetes/issues/93210), open a job-level issue targeted at the sig

2. If the whole job is failing (timeout, or failure to start/setup, or a majority of the tests in the job failing), open a job-level issue

3. Otherwise, open an issue for each SIG that has failing tests in the job.

I still think SIG/Job-level issues are as granular as ci-signal should go. SIG/component owners have the expertise to recognize when several distinct test failures have the same root cause, and can spawn/assign subissues if needed more effectively than CI signal folks.

hasheddan:
> I still think SIG/Job-level issues are as granular as ci-signal should go. SIG/component owners have the expertise to recognize when several distinct test failures have the same root cause, and can spawn/assign subissues if needed more effectively than CI signal folks. Asking CI signal helpers to open per-test/per-job issues is likely to result in wasted effort on their part and on the part of the assignees

I definitely see your point here, but there have been numerous instances where SIGs have specifically expressed frustration with issues being Job-level issues being opened (ref: https://github.com/kubernetes/kubernetes/issues/90955#issuecomment-662618915, others I can't find right now 😅 ). I am hoping with the CI Signal subproject that we are able to have more folks around with a little more granular context. Maybe the CI Signal team here should still focus on broader health though? That being said, it feels like these broad issues end up just creating noise.

liggitt:
In the spirit of holding SIGs accountable for their tests, I would tend to think that SIG/component owners are responsible for their tests passing in a job. If they want to create sub-issues tracking per-test fixes, that's totally fine, but that seems like their responsibility, not CI signal.

I also think 5 issues opened instead of https://github.com/kubernetes/kubernetes/issues/93485 would have been unhelpful ... it's hard to come up with mechanical guidance for when per-test issues are needed and when several stem from a similar issue

liggitt:
the main thing I'd like to accomplish here is to

1) make it clear who owns fixing the issue... scoping to SIG-level does that... they can delegate to subprojects if they want
2) make it clear what the criteria is for closing the issue... scoping to the SIG-owned tests does that too

justaugustus:
Adding @kubernetes/sig-scalability for input, as @wojtek-t mentioned this on an email thread: https://groups.google.com/d/msg/kubernetes-sig-release/8Ycc5pFZ000/42RqTYSwAgAJ

wojtek-t:
> In the spirit of holding SIGs accountable for their tests, I would tend to think that SIG/component owners are responsible for their tests passing in a job. If they want to create sub-issues tracking per-test fixes, that's totally fine, but that seems like their responsibility, not CI signal.

> I also think 5 issues opened instead of kubernetes/kubernetes#93485 would have been unhelpful ... it's hard to come up with mechanical guidance for when per-test issues are needed and when several stem from a similar problem

@liggitt 
It's probably case by case, but in case of SIG Scalability-owned jobs, it actually was pretty frustrating and counter-productive to have per-job issues. There are number of cases where one test fail on day1, people were debugging that over the next few days, than a second test failed, it was mentioned on the same issue, people started looking into it, other people are looking into previous one, both were pasting updates and there was complete mess impossible to follow.
And very very often, those were completely independent issues.

This is especially confusing, when those are flakes...


I agree that maybe the currently proposed procedure might be too simple, but I'm not a fan of what you're proposing actually.
To meet somewhere in the middle maybe something like:

1. If there is small number of tests failing (maybe the boundary is <3) - open issues targeting those individual tests
2. Otherwise, open job-level issue


BTW - in my experience, I found that issues targeting individual tests are much quicker fixed (or at the very least picked up and someone starts looking into them) than any of the job-oriented ones, e.g.:
https://github.com/kubernetes/kubernetes/issues/93424
https://github.com/kubernetes/kubernetes/issues/93341

liggitt:
> If there is small number of tests failing (maybe the boundary is <3) - open issues targeting those individual tests

Individual subissues per test where it makes sense is fine, but I would expect the SIG owner to drive creation of those, not the CI signal sig-release folks.

liggitt:
The signal sig-release is paying attention to is job-level. Making sure issues they open are targeted to a particular SIG for ownership seems like the right level of granularity to me.

wojtek-t:
But if there is single test failing, what's the purpose of creating a job-level issue, if the immediate thing to do later is to re-purpose it to test-oriented?

hasheddan:
I am happy to do whatever folks view as the most helpful, but FWIW as a CI Signal team member I actually prefer more granular issues that describe a test failure rather than a job. It is hard to know when to open and close issues for jobs because they may be healthy for a day or two then experience a new issue. I would prefer to have test-level issues that mention all the jobs they are affecting. I was attempting to make that even more apparent here by putting the relevant job in `()` in the issue title in this new format.

One other more extreme option would be to add labels for blocking jobs and then create issues for test-level failures and label them with blocking jobs that they are affecting. This would help with filtering and the cases where a test is affecting multiple jobs. I realize this would be a bigger change than anything else proposed here, so happy to table it, but that ultimately would be my preference. It would also make all SIGs more aware of how their tests affected release status.

liggitt:
> I also think 5 issues opened instead of kubernetes/kubernetes#93485 would have been unhelpful ... it's hard to come up with mechanical guidance for when per-test issues are needed and when several stem from a similar problem

That is still my opinion. If it's a toss-up whether individual or single issues are more desired  in general, I would lean towards the less toilsome option, and let subject-matter experts spawn per test issues where helpful

jkaniuk:
> That is still my opinion. If it's a toss-up whether individual or single issues are more desired in general, I would lean towards the less toilsome option, and let subject-matter experts spawn per test issues where helpful

From correctness tests perspective (especially scalability), job-level issues are much more toilsome, as all correctness failures need to be sub-issued again and there will be a mess of comments in both threads to follow. Rules for closing job level issue in such cases are unclear which leads to additional problems.

I like an idea to switch between test level and job level issues if there are more than, let's say, 2 tests failing.

liggitt:
In my ideal world, SIGs would just use the test grids and dashboards and wouldn't need a human opening issues for them at all. I think the CI-signal-spawned issues are a necessary concession to the lack of visibility to failures in omnibus jobs, but I really want to put the brunt of tracking/fixing the issues on the owning SIGs

wojtek-t:
Actually - in my experience reopening an issue was almost never a good idea. I would much rather prefer to have 5 issues opened and dedup them myself, than to have 3 different things be linked against the same issue.

Can we say that "never reopen an issue - always open a new one"?

liggitt:
> Actually - in my experience reopening an issue was almost never a good idea. I would much rather prefer to have 5 issues opened and dedup them myself, than to have 3 different things be linked against the same issue.

agree, reopening restarts conversations that are often no longer relevant

alejandrox1:
friendly ping @hasheddan

hasheddan:
Sorry for the (very) long delay on this one folks. I am ramping back up and have circled back to incorporate feedback and get this wrapped up. @alejandrox1 @liggitt @wojtek-t could you take another look?

release-team/role-handbooks/ci-signal/README.md
```diff
@@ -143,6 +130,39 @@ Jobs on the master-informing and release-x.y-informing dashboards require more i
...
 For more detailed information about what's on these dashboards, see [Release Blocking Jobs](/release-blocking-jobs.md) documentation.
 
+## Opening Issues
+
+The appropriate response to a failing or flaking job on either a blocking or informing dashboard is to open an issue. The primary goal of opening an issue is to track the work that needs to be done to bring the job back to a healthy status while providing accurate signal. The format for a helpful and informative issue may vary depending on the type of job, the test(s) that are failing for the job, and the responsible parties for the job.
+
+The decision tree below can be followed to make sure that you are opening an issue that is most effective for triaging, tracking, and resolving a failing or flaking job.
+
+_All instances of "failing" below can be replaced with "flaking" when appropriate._
+
```
liggitt, line 140: the content is a good improvement. the number/bullets aren't indenting quite right... might want to check the rendered output

hasheddan: @liggitt thanks for catching that! Should be fixed now :+1:

justaugustus:
Let's ship this and iterate as necessary.
Thanks for working this up, Dan!

/lgtm
/approve
cc: @kubernetes/ci-signal

PR #1230, opened 2020-09-15
Fix filename

jeremyrickard:
#### What type of PR is this:
/kind cleanup

This PR updates the 1.20 `release-team.md` file to be `release_team.md` to follow the pattern in previous releases to make it easier to jump from one release team to another when reviewing team membership. The old file is updated with a link to the correct file, in case anyone uses the old link (it was in an email). We can remove this file at some later point in the release when we have cleaned up all references. 

Signed-off-by: Jeremy Rickard <rickardj@vmware.com>


hasheddan (approved):
/lgtm

jrsapi (approved):

PR #1242, opened 2020-09-16
patches: update patch release schedule for 1.19/1.18/1.17

cpanato:
#### What type of PR is this:
/kind documentation

Update patch release page to add new the patch schedule for 1.19 / 1.18 and 1.17

None

/assign @tpepper @justaugustus 
/cc @kubernetes/release-managers


xmudrii (approved):
/lgtm
/hold
for the assigned folks to take a look

feiskyer (commented):
/lgtm
/approve

releases/patch-releases.md
```diff
@@ -59,47 +59,57 @@ releases may also occur in between these.
...
 ### 1.19
 
-Next patch release is **1.19.1**.
+Next patch release is **1.19.3**
 
-| Patch Release | Cherry-picks deadline | Target date |
-| --- | --- | --- |
-| 1.19.2 | 2020-09-11 | 2020-09-16 |
-| 1.19.1 | 2020-09-04 | 2020-09-09 |
+End of Life for **1.19** is **2021-09-30**
```
hasheddan, line 64: Is 1.19 supposed to have EOL while others are TBD?

cpanato: Tim added that in the previous Pr, but i agree we should define the EOL for others

hasheddan: Gotcha 👍 also was table formatting change intentional?

releases/patch-releases.md
```diff
@@ -59,47 +59,57 @@ releases may also occur in between these.
...
-| 1.18.5 | 2020-06-25 | 2020-06-26 |
-| 1.18.4 | 2020-06-12 | 2020-06-17 |
-| 1.18.3 | 2020-05-15 | 2020-05-20 |
-| 1.18.2 | 2020-04-13 | 2020-04-16 |
-| 1.18.1 | 2020-04-06 | 2020-04-08 |
+End of Life for **1.18** is **TBD**
+
+| PATCH RELEASE | CHERRY PICK DEADLINE | TARGET DATE |
+|--- |--- |--- |
+| 1.18.10       | 2020-10-09           | 2020-10-14  |
```
tpepper, line 80: Was this an automated formatting change from your editor, eg: in response to the column headers getting wider?  If so could that be disabled in your editor and the proposed formatting changes put in a separate PR?

I'm partly ok with the change in formatting as it feels more readable, but that should be a separate patch and we'd need to have some enforcing of the format so that somebody else's tool doesn't remove all the trailing spaces.

Or...was this partly the markdown generated from the yaml file plus some trailing space deleted from the 1.17 section cells?  If so...for a separate PR we should insert a non-printing comment in the document here saying "don't edit this file's markdown tables directly rather edit the yaml, run the generator, move the generated output into this file" (or better really commit to using the generator and update the generator to do that edit).

The goal of going into all the detail around formatting is to achieve a point where a PR like this is trivially and quickly reviewed because it has a +6 -3 diffstat instead of +60 −41 diffstat.

releases/patch-releases.md
```diff
@@ -59,47 +59,57 @@ releases may also occur in between these.
...
+| 1.17.11       | N/A | 2020-08-13  |
+| 1.17.10       | 2020-08-07 | 2020-08-12  |
+| 1.17.9        | 2020-07-10 | 2020-07-15  |
+| 1.17.8        | 2020-06-25 | 2020-06-26  |
+| 1.17.7        | 2020-06-12 | 2020-06-17  |
+| 1.17.6        | 2020-05-15 | 2020-05-20  |
+| 1.17.5        | 2020-04-13 | 2020-04-16  |
+| 1.17.4        | 2020-03-09 | 2020-03-12  |
+| 1.17.3        | 2020-02-07 | 2020-02-11  |
+| 1.17.2        | [No-op release](https://groups.google.com/d/topic/kubernetes-dev/Mhpx-loSBns/discussion) | 2020-01-21  |
```
tpepper, line 110: We may need to not allow variable length table cell content as this will lead to tooling adding/removing spaces in all the other cells complicating diffs and reviews.  A simple "N/A" should suffice, imho.  There are ways to get at what happened if somebody is curious.

tpepper (commented):
I'm not crazy about all the formatting change instead of just the dates changing.  But I can accept this PR as is if we keep going and commit to a format that is tool generated and we're all only using the tool.

releases/patch-releases.md
cpanato: I did some formating, to make nicer, but i can follow the one that already exists

cpanato: also i can use the one that the tool outputs

PR #1266, opened 2020-10-01
Clarify instructions to fetch all upstream branches

kcmartin:
Clarify instructions to state fetch all branches from upstream website repo, not just master branch

<!--  Thanks for sending a pull request!  Here are some tips for you:

#### What type of PR is this:

/kind documentation

Clarify instructions to fetch all branches from upstream website repo, not just master branch

None

### DIFF
--- a/release-team/role-handbooks/docs/README.md
+++ b/release-team/role-handbooks/docs/README.md
@@ -352,7 +352,7 @@ To merge `master` into `dev-[future release]` on your local fork:
 # Step 0 (if you don't already have a remote called "upstream")
 git remote add upstream https://github.com/kubernetes/website.git
 # Step 1
-git fetch upstream master
+git fetch upstream #grab all branches from upstream repo
 # Step 2
 git checkout --track upstream/dev-[future release]
 # Step 3
--- a/release-team/role-handbooks/docs/README.md
+++ b/release-team/role-handbooks/docs/README.md
@@ -352,21 +352,23 @@ To merge `master` into `dev-[future release]` on your local fork:
 # Step 0 (if you don't already have a remote called "upstream")
 git remote add upstream https://github.com/kubernetes/website.git
 # Step 1
-git fetch upstream #grab all branches from upstream repo
+git fetch upstream master
 # Step 2
-git checkout --track upstream/dev-[future release]
+git fetch upstream dev-[future release]
 # Step 3
+git checkout --track upstream/dev-[future release]
+# Step 4
 git pull --ff-only # make sure you're up to date
-# Step 4. You might see merge conflicts at this point.
+# Step 5 You might see merge conflicts at this point.
 git merge upstream/master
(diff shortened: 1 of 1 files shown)


release-team/role-handbooks/docs/README.md
```diff
@@ -352,7 +352,7 @@ To merge `master` into `dev-[future release]` on your local fork:
 # Step 0 (if you don't already have a remote called "upstream")
 git remote add upstream https://github.com/kubernetes/website.git
 # Step 1
-git fetch upstream master
+git fetch upstream #grab all branches from upstream repo
```
savitharaghunathan, line 355: Since we are pulling the changes from only `master` branch to keep `dev-[future release]` in sync, I guess fetching just `master` branch would suffice here and avoid fetching all other branches and associated changes.

annajung: We actually need to pull both `master` and `dev-[future release]`, so I don't think we want to pull all branches but add a step to fetch `dev-[future release]`.

@savitharaghunathan @kcmartin What do you think?

kcmartin: Hi-- the edit was prompted by an error I got following the steps in the role handbook:

`git checkout --track upstream/dev-1.20
fatal: 'upstream/dev-1.20' is not a commit and a branch 'dev-1.20' cannot be created from it`

After I used `git fetch upstream`, to fetch all the branches, I was able to complete the steps, hence the edit. I'm open to @annajung 's suggestion, or whatever makes sense, thanks!

kcmartin: Maybe we can fetch `master` and `dev-[future-release]` in just one step?

annajung: That's a great idea. I'm not sure if `git fetch` allows multiple branches? We should take a look and see! :) 

But, I don't think there's anything wrong with adding an additional command `git fetch upstream dev-[future release]` below `master` one. Or we can also do `git fetch upstream master && git fetch upstream dev-[future release]` to make it one-liner if git fetch doesn't allow multiple branches. 

Let's follow up on this and see if we can find anything, else I think adding extra command would be sufficient! 😀

savitharaghunathan: Hi @kcmartin, Thank you for adding the error message, now I remember going through that and had to fetch dev-branch. I use to alternate between fetching everything and just fetching what is required. My apologies for the review comment. I see no harm now in adding an additional step or merging @kcmartin's suggestion.

kcmartin:
What do you think @annajung and @savitharaghunathan ?
I added the additional step for fetching the dev branch.
Thank you both so much for the thoughtful review comments!

annajung:
/lgtm

I'll leave approval to @savitharaghunathan

savitharaghunathan (commented):
/lgtm
/approve

PR #1274, opened 2020-10-09
add 1.20 exceptions.yaml

kikisdeliveryservice:
Starting the exceptions.yaml file for 1.20 Release.

### DIFF
--- a/releases/release-1.20/exceptions.yaml
+++ b/releases/release-1.20/exceptions.yaml
@@ -1,0 +1,45 @@
+# Exceptions requested in 1.20
+# Google Group: https://groups.google.com/forum/#!forum/kubernetes-sig-release
+# Release Team Lead: Jeremy Rickard ([@jeremyrickard](https://www.github.com/jeremyrickard))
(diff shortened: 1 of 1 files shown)


saschagrunert (approved):

releases/release-1.20/exceptions.yaml
```diff
@@ -0,0 +1,24 @@
+# Exceptions requested in 1.20
+# Google Group: https://groups.google.com/forum/#!forum/kubernetes-sig-release
+# Release Team Lead: Kirsten Garrison (@kikisdeliveryservice)
```
justaugustus, line 3: This should match the RT Leads and Lead Shadows.

Example: https://github.com/kubernetes/sig-release/blob/db606943bb342c1ab55cbc4b7809a6e7cb6ebd6d/releases/release-1.19/exceptions.yaml#L3-L4

justaugustus (changes requested):
@kikisdeliveryservice -- I think there are more exceptions approved from this cycle since this PR was created.

Outside of that, one nit on the Leads listing.

releases/release-1.20/exceptions.yaml
kikisdeliveryservice: Oh gosh DUH. Was rushing will update. 🤦

kikisdeliveryservice:
OK finally updated :smile: 

@justaugustus PTAL

/hold cancel

PR #1285, opened 2020-10-13
update exceptions.yaml to incl exceptions 10-12 through 10-13

kikisdeliveryservice:
Adding/updating 2 exceptions approved October 12&13th.


hasheddan (approved):
/lgtm

Issue #912, opened 2019-12-12
Ensure all release team handbooks include an item for holding an initial team meeting with their shadows


karenhchu:
We need to add to the release team lead handbook a note holding an initial team meeting with their shadows in order to introduce everyone, their collective tasks, and upcoming timeline. This will better foster a team atmosphere and allow members to help/cover for each other accordingly.

justaugustus:
Great idea, @karenhchu!
I think this should be listed in every handbook.

LappleApple:
In our release team process-improvement prioritisation sessions a few folks mentioned taking a closer look at the general RT onboarding process and make some adjustments. Maybe we can try to do that broader review of onboarding, create an initiative for it, and include this item as a to-do of that initiative?

karenhchu:
Sure thing. Should I leave this open or close it since it'll be addressed in newer efforts @LappleApple?

lauriapple2:
Hey @karenhchu: I'd say let's leave it open for a bit longer, because this week I am going to work a bit on organizing our thoughts on onboarding and driving inclusivity.

palnabarun:
Based on the above discussions and discussions over the Release Team meetings, I have retitled the issue to mention the latest quo.

Also, putting out a checklist for each team handbook.
- [ ] Bug Triage
- [ ] CI Signal 
- [ ] Communications 
- [ ] Docs
- [ ] Enhancements
- [ ] Release Notes

PS: This is a nice task to be done by the shadows of each team.

cc: @kubernetes/release-team 

/assign

palnabarun:
Enhancements Handbook would be updated through https://github.com/kubernetes/sig-release/pull/1304/commits/df174e62dace7c7fbb5e64dc8647c90539c182f8#diff-c853db51181791969e437e9f720a7ef21fbcf460f1eebda80b8ae6a7e2673ba8R96 :)

annajung:
Docs handbook already has a section to do this and can be marked off as completed
https://github.com/kubernetes/sig-release/blob/master/release-team/role-handbooks/docs/Release-Timeline.md#meet-with-shadows

palnabarun:
Thank you @annajung for the update. 😃

jeremyrickard:
Bug Triage updated in https://github.com/kubernetes/sig-release/pull/1345

jeremyrickard:
ANNNND comms is updated ☝️ . Going to close this issue now!

jeremyrickard:
/close 

[...]

PR #1308, opened 2020-10-27
Update CI Signal handbook to include holding an onboarding meeting for shadows

thejoycekung:
#### What type of PR is this:
/kind documentation

To make sure any future CI Signal leads are aware they should be holding an onboarding meeting for shadows

Relates to #912 

/cc @RobertKielty


hasheddan (approved):
/lgtm

Thanks @thejoycekung!

Issue #1337, opened 2020-11-15
[Umbrella] Formalize supported release platforms


hasheddan:
We currently build and release k/k artifacts for a number of os / arch platforms (listed below). However, there is no formal process for adding a new platform, retiring an existing platform, determining level of support for an existing platform, or measuring test coverage for an existing platform. The currently supported platforms are listed below: 

```
amd64-linux
arm64-linux
arm-linux
amd64-darwin
386-linux
ppc64le-linux
s390x-linux
386-windows
amd64-windows
```

This umbrella issue tracks the steps to accomplish the following objectives. It will be updated and linked to more granular issues and PRs.
- Make it abundantly clear to end users which platforms are supported and their level of support
    - [x] Update document in k/sig-release with currently supported platforms @hasheddan 
    - [ ] #1720
    - [ ] #1721
    - [x] Define "tiers" of support for a platform @saschagrunert 
- Document current blocking / informing jobs @hasheddan 
- Verify that we are in fact publishing artifacts for every platform we support
    - [ ] Add step to krel stage / release to verify existence of artifacts for all supported platforms @hasheddan 
- Define steps for introducing a new platform
    - [x] Add guide for shipping alternative platforms (https://github.com/kubernetes/community/pull/5300) @saschagrunert 
- Define steps for retiring and existing platform
    - Likely requires a broader conversation about whether we can drop support for a platform (i.e. deprecation / removal cycle)
- Define metrics around test coverage for supported platforms (i.e. make our level of support somewhat quantifiable)
   - [ ] #1722

/kind documentation
/area release-eng
/priority important-soon
/assign
/assign @saschagrunert

saschagrunert:
Looks good, I'll continue with the Tiers on kubernetes/community#5300 got merged.

hasheddan:
Found this google groups thread and want to loop in @BenTheElder and @johnbelamaric to let y'all know this effort is underway. I am planning on attending SIG-Arch meeting this Thursday https://groups.google.com/g/kubernetes-sig-architecture/c/eHBFgfd6Qxg

hasheddan:
Slides from presentation to sig-arch: https://docs.google.com/presentation/d/1bctKOljDHwCB6115pphPO3VQ4PiLHcOqYiBA5IXorTE/edit?usp=sharing

saschagrunert:
[...]

PR #1352, opened 2020-11-22
Add supported platforms document

hasheddan:
#### What type of PR is this:

/kind documentation

Adds the initial documentation on supported platforms for each
Kubernetes component. This document will continue to be iterated upon as
procedures for supporting / dropping support for platforms becomes more
defined. It will also be updated with tiers of support and current test
coverage.

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>

xref: https://github.com/kubernetes/sig-release/issues/1337

/assign @saschagrunert @justaugustus 
/area release-eng

### DIFF
--- a/supported-platforms.md
+++ b/supported-platforms.md
@@ -1,0 +1,20 @@
+# Supported Platforms
+
+A Kubernetes release consists of a set of Go binaries for various Operating Systems and CPU architectures. Platforms are identified using the Go `<GOOS>-<GOARCH>` parlance.
+
+## Released Binaries
+
+Kubernetes currently releases binaries for the following components on the listed platforms:
+
+|   Component / Platform  | darwin-amd64 | linux-386 | linux-amd64 | linux-arm | linux-arm64 | linux-ppc64le | linux-s390x | windows-386 | windows-amd64 |
+|:-----------------------:|:------------:|:---------:|:-----------:|:---------:|:-----------:|:-------------:|:-----------:|:-----------:|:-------------:|
+| apiextensions-apiserver |              |           |      X      |     X     |      X      |       X       |      X      |             |               |
+|     kube-aggregator     |              |           |      X      |     X     |      X      |       X       |      X      |             |               |
+|      kube-apiserver     |              |           |      X      |     X     |      X      |       X       |      X      |             |               |
+| kube-controller-manager |              |           |      X      |     X     |      X      |       X       |      X      |             |               |
+|        kube-proxy       |              |           |      X      |     X     |      X      |       X       |      X      |             |       X       |
+|      kube-scheduler     |              |           |      X      |     X     |      X      |       X       |      X      |             |               |
+|         kubeadm         |              |           |      X      |     X     |      X      |       X       |      X      |             |       X       |
+|         kubectl         |       X      |     X     |      X      |     X     |      X      |       X       |      X      |      X      |       X       |
+|         kubelet         |              |           |      X      |     X     |      X      |       X       |      X      |             |       X       |
+|         mounter         |              |           |      X      |     X     |      X      |       X       |      X      |             |               |


justaugustus (changes requested):
@hasheddan -- Very awesome!

Two things:
- As we start to draw the line between what is released and what is supported, I think it's important to put a disclaimer here, something in the way of "We test on linux/amd64 and that's the only thing we're 'really sure' about. Everything else is best effort."
- Can you break these out into two tables: server and client

dims:
/hold

for reviews! :)

supported-platforms.md
```diff
@@ -0,0 +1,20 @@
+# Supported Platforms
```
saschagrunert, line 1: Do we want to put this into `/release-engineering`? And link it from the `README.md` there?

supported-platforms.md
```diff
@@ -0,0 +1,20 @@
...
+
+A Kubernetes release consists of a set of Go binaries for various Operating Systems and CPU architectures. Platforms are identified using the Go `<GOOS>-<GOARCH>` parlance.
```
saschagrunert, line 3:
```suggestion
A Kubernetes release consists of a set of Go binaries and container images for various Operating Systems and CPU architectures. Platforms are identified using the Go `<GOOS>-<GOARCH>` parlance.
```

supported-platforms.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Released Binaries
```
saschagrunert, line 5: I think we can map this 1:1 to the container images, right? I guess we just have to mention this fact here, too.

supported-platforms.md
```diff
@@ -0,0 +1,20 @@
...
+| kube-controller-manager |              |           |      X      |     X     |      X      |       X       |      X      |             |               |
+|        kube-proxy       |              |           |      X      |     X     |      X      |       X       |      X      |             |       X       |
+|      kube-scheduler     |              |           |      X      |     X     |      X      |       X       |      X      |             |               |
+|         kubeadm         |              |           |      X      |     X     |      X      |       X       |      X      |             |       X       |
+|         kubectl         |       X      |     X     |      X      |     X     |      X      |       X       |      X      |      X      |       X       |
+|         kubelet         |              |           |      X      |     X     |      X      |       X       |      X      |             |       X       |
+|         mounter         |              |           |      X      |     X     |      X      |       X       |      X      |             |               |
```
saschagrunert, line 20: As a follow-up, my idea would be that we put this matrix in a map in k/release and verify it during the release cut. WDYT?

saschagrunert (commented):
Thank you! Just some thoughts on container images, otherwise LGTM :point_down:

supported-platforms.md
hasheddan: @saschagrunert definitely :) I have an item on #1337 to do that :+1:

supported-platforms.md
hasheddan: Yep that sounds good :+1:

supported-platforms.md
hasheddan: So I was thinking about that, but I think as this document continues to mature, it actually is generally applicable to sig-release, similar to how the release-blocking / release-informing doc is at root of repo.

supported-platforms.md
hasheddan: @saschagrunert in thinking about this more, the container images we publish don't really map 1:1 here, so I think it makes sense to maybe follow-up with some additional documentation on that

supported-platforms.md
spiffxp: There are also various scripts and manifests, looking at the contents of e.g. https://github.com/kubernetes/kubernetes/releases/download/v1.18.12/kubernetes.tar.gz

hasheddan: @spiffxp yep, this is not meant to cover _every artifact_ produced, just the supported platforms. That being said, I do think we should document those as well.

hasheddan:
So I completely missed that an excellent document that describes what I have here as well as much more already exists (https://github.com/kubernetes/sig-release/blob/master/release-engineering/artifacts.md). I am going to close this out and do some work on cleaning up the existing doc :+1: 

/close

PR #1360, opened 2020-11-30
Add platforms documentation

saschagrunert:
#### What type of PR is this:

/kind documentation

This adds the documentation about supported Kubernetes platforms to our
repository.

Refers to #1337 
None

### DIFF
--- a/release-engineering/platforms.md
+++ b/release-engineering/platforms.md
@@ -1,0 +1,68 @@
+# Platform Builds
+
+The Kubernetes project's development process produces
+[artifacts](./artifacts.md) for different architectures and operating systems.
+We consider the combination of architecture (`GOARCH`) and operating system
+(`GOOS`) as "platforms". Target of this document is to outline different
+categories of platforms as well as guiding through their graduation criteria.
+
+## Tiers
+
+Build and release support for different platforms' artifacts are organized into
+three tiers, whereas each comes with a different set of guarantees. Tiers can be
+scoped to single binaries or a subset of them. This means for example that we
+can provide Tier 1 support for client binaries, even if the server binaries do
+not exist at all.
+
+### Tier 1
+
+Tier 1 platforms are considered as "expected to work". To achieve this, they
+have to fulfill the following criteria:
+
+- Official binary releases are provided for the platform. This includes
+  container images as well as deb and rpm packages. Building the artifacts is
+  integrated in the release process.
+- Continuous Integration is set up to run tests for the platform. Necessary
+  tests are defined by SIG Release and usually correspond to the
+  [`blocking`](https://testgrid.k8s.io/sig-release-master-blocking) and
+  [`informing`](https://testgrid.k8s.io/sig-release-master-informing) testgrid
+  dashboards.
+- Documentation about the usage of artifacts for the platform is available.
+
+### Tier 2
+
+Tier 2 platforms are considered as "expected to build". To achieve this, they
+have to fulfill the following criteria:
+
+- Official binary releases are provided for the platform. Building the artifacts
+  is integrated in the release process.
+- Automated testing is not or only partially setup.
+
+It may be possible that single features are not available for a certain
+platform.
+
+### Tier 3
+
+Tier 3 platforms are those which the Kubernetes codebase has functional support
+for, but which are not built or tested automatically. This means the binary
+artifacts may not work as intended.
+
+- Official builds are not available.
+- Automated testing is not setup.
+
+## Currently available Kubernetes platforms
+
+The following table defines the current setup of available Kubernetes platforms:
+
+| Platform        |       Tier 1       |       Tier 2       |       Tier 3       |
+| --------------- | :----------------: | :----------------: | :----------------: |
+| `amd64-linux`   | :heavy_check_mark: |                    |                    |
+| `arm64-linux`   |                    | :heavy_check_mark: |                    |
+| `arm-linux`     |                    | :heavy_check_mark: |                    |
+| `amd64-darwin`  |                    | :heavy_check_mark: |                    |
+| `386-linux`     |                    | :heavy_check_mark: |                    |
+| `ppc64le-linux` |                    | :heavy_check_mark: |                    |
+| `s390x-linux`   |                    | :heavy_check_mark: |                    |
+| `386-windows`   |                    | :heavy_check_mark: |                    |
+| `amd64-windows` |                    | :heavy_check_mark: |                    |
+| otherwise       |                    |                    | :heavy_check_mark: |


release-engineering/platforms.md
```diff
@@ -0,0 +1,65 @@
...
+- **Any** official binary release is provided for the platform. Building the
+  artifacts is integrated in the release process.
+- Automated testing is not or only partially setup.
+
+It may be possible that single features are not available for a certain
+platform.
+
+### Tier 3
+
+Tier 3 platforms are those which the Kubernetes codebase has support for, but
```
hasheddan, line 43: What do you think "has support for" would encompass here? As it currently stands I think it could include _any_ platform, regardless of if there has been any effort to build Kubernetes for it. If that is the case, then I'm wondering how useful having a third tier is vs just calling these platforms unsupported. I would like to build out some tooling to be able to do experimental builds for new platforms we want to introduce. Maybe that could be tier 3?

release-engineering/platforms.md
```diff
@@ -0,0 +1,65 @@
...
+  [`informing`](https://testgrid.k8s.io/sig-release-master-informing) testgrid
+  dashboards.
+- Documentation about the usage of artifacts for the platform is available.
+
+### Tier 2
+
+Tier 2 platforms are considered as "guaranteed to build". To achieve this, they
+have to fulfil the following criteria:
+
+- **Any** official binary release is provided for the platform. Building the
```
hasheddan, line 34: This is kind of tricky because we currently release artifacts that are not well tested or tested at all, but in the future it would be great if we "released" only Tier 1 artifacts, and published Tier 2 separately. However, I think we have to stick with that is defined here for now given our current state.

hasheddan (commented):
Thanks for adding this @saschagrunert! Left a few comments below, but mostly looks good to me. Let's allow some others to weigh in as well :)

/hold

release-engineering/platforms.md
saschagrunert: I see the point. There are some platforms which do not even support golang, so those would be definitely "unsupported". You're right, every valid golang platform which is not in tier 1 or 2 would be tier 3. It might build with version X, but since there is no CI in place it might break with future commits.

release-engineering/platforms.md
```diff
@@ -0,0 +1,65 @@
...
+comes with a different set of guarantees.
+
+### Tier 1
+
+Tier 1 platforms are considered as "guaranteed to work". To achieve this, they
+have to fulfil the following criteria:
+
+- **All** official binary releases are provided for the platform. This includes
+  container images as well as deb and rpm packages. Building the artifacts is
+  integrated in the release process.
```
hasheddan, line 21: I am wondering if maybe support levels should be "per-binary". For instance if we thoroughly test `kubectl` on `darwin-amd64` it would technically still fall into Tier 2 by this definition.

release-engineering/platforms.md
hasheddan: Sounds good 👍

release-engineering/platforms.md
mkorbi: @hasheddan do we track this open flanks already?

release-engineering/platforms.md
saschagrunert: Alright, I changed the doc to outline that we can scope Tiers to single binaries. Let me know what you think.

hasheddan (approved):
@saschagrunert thanks for the update!

/lgtm

release-engineering/platforms.md
```diff
@@ -0,0 +1,67 @@
+# Supported Platforms
```
tpepper, line 1: Could this be "Platform Builds"?

The support word is confusing.  The paragraph following this title doesn't define support.  For the context of this document it feels like "support" is meaning something around "how SIG Release and its Release Engineering subproject stand behind the artifacts they build", especially because of the repo and subdirectory in which the file sits and how it starts with "The Kubernetes release process produces artifacts..."

But it's easy to read something bigger as implied, eg: "The Kubernetes project's development process supports artifacts..."  This is a subtle but important distinction in terms of correctly setting expectations on the reader's side.

release-engineering/platforms.md
```diff
@@ -0,0 +1,67 @@
...
+## Tiers
+
+Support for different platforms are organized into three tiers, whereas each
+comes with a different set of guarantees. Tiers can be scoped to single
+binaries or a subset of them. This means for example that we can provide Tier 1
+support for client binaries, even if the server binaries do not exist at all.
+
+### Tier 1
+
+Tier 1 platforms are considered as "guaranteed to work". To achieve this, they
```
tpepper, line 18: Suggest: "expected to work"

release-engineering/platforms.md
```diff
@@ -0,0 +1,67 @@
...
+have to fulfil the following criteria:
```
tpepper, line 19: typo: fulfill

release-engineering/platforms.md
```diff
@@ -0,0 +1,67 @@
...
+- Continuous Integration is set up to run tests for the platform. Necessary
+  tests are defined by SIG Release and usually correspond to the
+  [`blocking`](https://testgrid.k8s.io/sig-release-master-blocking) and
+  [`informing`](https://testgrid.k8s.io/sig-release-master-informing) testgrid
+  dashboards.
+- Documentation about the usage of artifacts for the platform is available.
+
+### Tier 2
+
+Tier 2 platforms are considered as "guaranteed to build". To achieve this, they
```
tpepper, line 33: "expected to build"

release-engineering/platforms.md
```diff
@@ -0,0 +1,67 @@
...
+- Official binary releases are provided for the platform. Building the artifacts
+  is integrated in the release process.
+- Automated testing is not or only partially setup.
+
+It may be possible that single features are not available for a certain
+platform.
+
+### Tier 3
+
+Tier 3 platforms are those which the Kubernetes codebase has support for, but
```
tpepper, line 45: suggest "codebase has functional support for"

release-engineering/platforms.md
```diff
@@ -0,0 +1,67 @@
...
+
+The Kubernetes release process produces [artifacts](./artifacts.md) for
+different architectures and operating systems. We consider the combination of
+architecture (`GOARCH`) and operating system (`GOOS`) as "platforms". Target of
+this document is to outline different categories of platforms as well as guiding
+through their graduation criteria.
+
+## Tiers
+
+Support for different platforms are organized into three tiers, whereas each
```
tpepper, line 11: suggest "Build and release support for different platforms' artifacts are..."

tpepper (changes requested):
I think as stated this document can give a reader a wrong impression of what is community support and the overall Kubernetes project's support stance, versus covering what it is that SIG Release does within that broader context.  I've suggested a few slight changes in wording which soften what I think reads as overstatements around the term "support".

release-engineering/platforms.md
saschagrunert: Changed as suggested :+1:

release-engineering/platforms.md
saschagrunert: Changed as suggested :+1:

release-engineering/platforms.md
saschagrunert: Changed as suggested :+1:

release-engineering/platforms.md
saschagrunert: Fixed

release-engineering/platforms.md
saschagrunert: Changed as suggested :+1:

release-engineering/platforms.md
saschagrunert: Yes I agree, changed as suggested.

saschagrunert:
> I think as stated this document can give a reader a wrong impression of what is community support and the overall Kubernetes project's support stance, versus covering what it is that SIG Release does within that broader context. I've suggested a few slight changes in wording which soften what I think reads as overstatements around the term "support".

Makes sense, I changed the document as suggested. Please take a look again.

hasheddan (approved):
Re-adding my LGTM :) thanks again @saschagrunert!

/lgtm

saschagrunert:
I think we addressed most concerns and we can lift the hold.
/hold cancel

PR #1367, opened 2020-12-01
releng: Update guidelines on promotiong RC images

xmudrii:
#### What type of PR is this:

/kind documentation

Update the image promotion handbook to add that alpha.0 images should be promoted when promoting rc.0 images.

Reference: https://kubernetes.slack.com/archives/CJH2GBF7Y/p1606823807291300?thread_ts=1606811420.282800&cid=CJH2GBF7Y

None.

None.

/assign @hasheddan @saschagrunert @cpanato 
cc: @kubernetes/release-engineering

### DIFF
--- a/release-engineering/role-handbooks/release-image-promotion.md
+++ b/release-engineering/role-handbooks/release-image-promotion.md
@@ -30,19 +30,18 @@ The images are promoted by using the `krel promote-images` command. This command
   - Otherwise, you can specify both GitHub username and fork name, such as `--fork=<your-github-username>/<k8s.io-fork-name>`
 - The `--tag` flag takes the version tag of the images that will be promoted. This flag can be specified multiple times if you're promoting images for multiple releases.
 
-Depending on the release you're cutting, you need to run the following command:
-
-- If you're cutting Alpha, Beta, or RC release, use the following `krel promote-images` command to promote images for the release you're cutting:
+Example command:
 
 ```shell
-krel promote-images -i --fork=<your-github-username> --tag=v1.20.0-beta.1
+krel promote-images -i --fork=<your-github-username> --tag=v1.20.0-rc.0 --tag=v1.21.0-alpha.0
 ```
 
-- If you're cutting a stable release (e.g. `v1.20.0`), you need to promote the images for the release you're cutting and for the next RC release (e.g. `v1.20.1-rc.0`). You can use the following command to do that:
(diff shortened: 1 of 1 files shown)


release-engineering/role-handbooks/release-image-promotion.md
```diff
@@ -32,12 +32,24 @@ The images are promoted by using the `krel promote-images` command. This command
...
+
+```shell
+krel promote-images -i --fork=<your-github-username> --tag=v1.20.0-rc.0 --tag=v1.21.0-alpha.0
+```
+
+- If you're cutting the subsequent RC releases (e.g. `v1.20.0-rc.1`), you only need to promote the images for the RC:
+
+```shell
+krel promote-images -i --fork=<your-github-username> --tag=v1.20.0-rc.1
+```
```
xmudrii, line 51: I believe that it should be like that, but is that the case actually? I just want to double-check to be sure.

hasheddan: This should be correct :)

release-engineering/role-handbooks/release-image-promotion.md
```diff
@@ -32,12 +32,24 @@ The images are promoted by using the `krel promote-images` command. This command
 
 Depending on the release you're cutting, you need to run the following command:
 
-- If you're cutting Alpha, Beta, or RC release, use the following `krel promote-images` command to promote images for the release you're cutting:
+- If you're cutting Alpha or Beta release, use the following `krel promote-images` command to promote images for the release you're cutting:
```
hasheddan, line 35: nitpick: do you think it would be worth generalizing this section and just saying "this is the command syntax", then following with notes about the special cases of first rc and stable versions requiring two tag args?

xmudrii: @hasheddan The command syntax is described in detail above. Let me see can I come up with something.

hasheddan: Right, I am just thinking we are pretty verbose here, which maybe is good? But maybe we could make it a little more succinct. Happy to go ahead and move forward if you prefer the more verbose approach :)

xmudrii: @hasheddan I've updated the PR, PTAL. This should be more concise and easier to read, but also (hopefully) verbose enough.

hasheddan (approved):
/lgtm

PR #1371, opened 2020-12-03
Release notes draft v1.20.0 rc.0

wilsonehusin:
#### What type of PR is this:

/kind documentation

#### What this PR does / why we need it:

Release notes for Kubernetes 1.20, based on `v1.20.0-rc.0` tag

#### Which issue(s) this PR fixes:

None

#### Special notes for your reviewer:

I will go through this again tomorrow with the community documentation guide in hand, too tired now and just want to get this out first :smile:

/cc @celestehorgan @tanjacky @soniasingla @JamesLaverack @puerco

PR #1373, opened 2020-12-03
Add release notes JSON draft

saschagrunert:
#### What type of PR is this:

/kind feature

We will use this JSON draft as artifact during the release process,
which allows consumers to get a machine-readable format of the notes.

The notes have been generated via:

```
> release-notes --format=json \
    --maps-from=$GOPATH/src/k8s.io/sig-release/releases/release-1.20/release-notes/maps \
    --start-rev=v1.19.0 \
    --end-rev=v1.20.0-rc.0
```

Refers to https://github.com/kubernetes/release/issues/1087 and https://github.com/kubernetes/release/pull/1815
/cc @wilsonehusin @puerco


puerco:
Should be fine when paired #1731, that is unless more edits happen in the coming days (but we should have automated this by then)
/lgtm

wilsonehusin:
oh interesting, I didn't know this workflow.

I made edits to the Markdown Draft PR directly on Markdown instead of krel because it was too many to go through sequentially when I already know rhe PR numbers to edit.

I'm guessing this command is running through the map instead of the markdown -- correct? if so, that means this JSON might be missing some entries which I added to Markdown manually 😐

saschagrunert:
> oh interesting, I didn't know this workflow.
> 
> I made edits to the Markdown Draft PR directly on Markdown instead of krel because it was too many to go through sequentially when I already know rhe PR numbers to edit.
> 
> I'm guessing this command is running through the map instead of the markdown -- correct? if so, that means this JSON might be missing some entries which I added to Markdown manually 😐

We always had a gap between the manual edited release notes and the generated JSON in the past, so I guess it’s fine. In the future we can use the maps to synchronize both versions directly.

wilsonehusin:
> Are you referring to the mapping workflow itself? Do we lack documentation about it?

JSON file :smile:

> We always had a gap between the manually edited release notes and the generated JSON in the past, so I guess it’s fine. In the future we can use the maps to synchronize both versions directly.

gotcha! LGTM

/hold cancel

wilsonehusin:
for the record, these are the missing notes which I manually added to #1371

- https://github.com/kubernetes/kubernetes/pull/91824
- https://github.com/kubernetes/kubernetes/pull/95856
- https://github.com/kubernetes/kubernetes/pull/95896
- https://github.com/kubernetes/kubernetes/pull/94879
- https://github.com/kubernetes/kubernetes/pull/95896

PR #1384, opened 2020-12-10
patches/schedule: add January 2021 patch schedule

cpanato:
#### What type of PR is this:

/kind documentation

patches/schedule: add January 2021 patch schedule

None

/assign @justaugustus @saschagrunert @hasheddan

### DIFF
--- a/releases/patch-releases.md
+++ b/releases/patch-releases.md
@@ -47,21 +47,31 @@ releases may also occur in between these.
 
 | Monthly Patch Release | Target date |
 | --- | --- |
-| December 2020 | 2020-12-09 |
 | January 2021 | 2021-01-13 |
 | February 2021 | 2021-02-17 |
 | March 2021 | 2021-03-17 |
 
 ## Detailed Release History for Active Branches
 
+### 1.20
+
+Next patch release is **1.20.1**
+
+End of Life for **1.20** is **2021-12-30**
+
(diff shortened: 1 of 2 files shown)


releases/patch-releases.md
```diff
@@ -47,21 +47,31 @@ releases may also occur in between these.
...
 
+### 1.20
+
+Next patch release is **1.20.1**
+
+End of Life for **1.20** is **2021-12-30**
+
+| PATCH RELEASE | CHERRY PICK DEADLINE | TARGET DATE |
+|--- |--- |--- |
+| 1.20.1        | 2021-01-08           | 2021-01-13  |
```
saschagrunert, line 64: I'm happy with the date and also wondering if we want to shift +1 week to ensure that most people are back from their holidays. On the other hand I do not expect many changes coming in during Christmas.

WDYT?
cc @justaugustus

cpanato: I was using the dates that were defined, but I agree that we can push a week

hasheddan: 1/13/21 feels like enough time to me for folks to get back from holiday, but bumping it out would be fine as well.

cpanato:
@hasheddan @saschagrunert @tpepper should we update the dates to `1/13/21` ? or keep them as is?

Issue #1404, opened 2021-01-07
Release Team Lead access for @palnabarun @kikisdeliveryservice @savitharaghunathan @bai


palnabarun:
### GitHub Username(s)

- @palnabarun 
- @kikisdeliveryservice 
- @savitharaghunathan 
- @bai 

### Release Cycle

Kubernetes 1.21

### Prerequisites

- [x] Ensure new Release Team Lead:
  - [x] Has joined the following mailing lists:
    - [kubernetes-sig-release](https://groups.google.com/forum/#!forum/kubernetes-sig-release)
    - [kubernetes-dev](https://groups.google.com/forum/#!forum/kubernetes-dev)
  - [x] Has joined the following Slack channels:
    - [#sig-release](https://kubernetes.slack.com/messages/C2C40FMNF)
    - [#release-management](https://kubernetes.slack.com/messages/CJH2GBF7Y)
  - [x] Is a [Kubernetes GitHub org member](https://github.com/kubernetes/community/blob/master/community-membership.md#member)
  - [x] Has completed the [CNCF Inclusive Speaker Orientation](https://training.linuxfoundation.org/training/inclusive-speaker-orientation/). Please drop in the certification ID in the comments once completed.
    - [x] Nabarun Pal
    - [x] Kirsten Garrison
    - [x] Savitha Raghunathan
    - [x] Vlad Gorodetsky

### Onboarding

- [x] Release Team Lead has agreed to abide by the guidelines set forth in the
  [Security Release Process](https://git.k8s.io/security/security-release-process.md), specifically the embargo on CVE communications.
  This must be done as an issue comment by the incoming Release Team Lead.
- [x] Update GitHub teams [(`kubernetes/org`)](https://git.k8s.io/org/config/kubernetes/sig-release/teams.yaml) ~> (https://github.com/kubernetes/org/pull/2405)
  - `milestone-maintainers`
  - `release-team`
  - `release-team-leads`
  - `sig-release`
- [x] Update `kubernetes/sig-release` `OWNERS` ~> (https://github.com/kubernetes/sig-release/pull/1407 and https://github.com/kubernetes/sig-release/pull/1405)
  - Release Team Lead and Shadows
    - Add an `approvers` entry in `releases/release-1.21/OWNERS`
  - **Release Team Lead only**
    - In `OWNERS_ALIASES`, add an entry in the following sections:
      - `release-team`
      - `release-team-lead-role`
[...]

PR #1407, opened 2021-01-08
release-1.21: update OWNER_ALIASES

palnabarun:
#### What type of PR is this:

/kind cleanup
/kind feature

Updates the following OWNER_ALIASES with the 1.21 team:
- release-team-lead-role
- ci-signal-role
- enhancements-role
- bug-triage-role
- docs-role
- release-notes-role
- communications-role

Thank you @lachie83 for being the Emeritus Advisor for 1.20 Release Team. ❤️ 

ref: https://github.com/kubernetes/sig-release/issues/1404

I noticed that 1.17 is yet to be EOL. So, kept the role lead entries for 1.17.

/assign
/sig release
/area release-team
/priority critical-urgent
/assign @justaugustus @saschagrunert @jeremyrickard @hasheddan @alejandrox1
/cc @kikisdeliveryservice @savitharaghunathan @bai

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -36,6 +36,7 @@ aliases:
     - jeremyrickard # subproject owner / 1.20 RT Lead
     - justaugustus # subproject owner
     - onlydole # subproject owner / 1.19 RT Lead
+    - palnabarun # subproject owner / 1.21 RT Lead
     - tpepper # subproject owner
   build-admins:
     - amwat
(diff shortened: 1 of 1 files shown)


OWNERS_ALIASES
```diff
@@ -45,36 +46,42 @@ aliases:
   release-team-lead-role:
     - guineveresaenger # 1.17 RT Lead
```
savitharaghunathan, line 47: Last time we had 4 entries including the current lead, are we retaining leads from 4 previous releases in addition to the current lead?

hasheddan (approved):
/lgtm

OWNERS_ALIASES
hasheddan: @savitharaghunathan I think we keep them as long as a release branch is still active

palnabarun: So, we were retaining folks from only the releases that are in-support. Since 1.17 is going to be EOL by end of this month, I am deferring the cleanup to later.

savitharaghunathan: Makes sense. Thank you @hasheddan & @palnabarun for the explanation.

PR #1409, opened 2021-01-08
update access reqs for enhancements handbook

kikisdeliveryservice:
Update enhancements handbook to remove Owners_aliases requirement (which is stale and milestone maintainer gives all needed permissions) and to add links to appropriate github teams for clarity.

### DIFF
--- a/release-team/role-handbooks/enhancements/README.md
+++ b/release-team/role-handbooks/enhancements/README.md
@@ -61,9 +61,8 @@ The shadows should be selected keeping in mind that one of them may eventually b
 
 Ensure that the previous Enhancements Lead has given (or facilitated getting) you access to:
 - GitHub teams
-  - enhancements (This group should be used for Enhancement Subproject related pinging only and not for Release Team Enhancements Group)
-  - milestone-maintainers
-- OWNERS_ALIASES (as `enhancements` in [kubernetes/enhancements][k/enhancements] repo)
+  - [enhancements](https://git.k8s.io/org/config/kubernetes/sig-architecture/teams.yaml) (This group should be used for Enhancement Subproject related pinging only and not for Release Team Enhancements Group)
+  - [milestone-maintainers](https://git.k8s.io/org/config/kubernetes/sig-release/teams.yaml)
 - Access to the previous Kubernetes Release Enhancements Tracking Sheet.
 
 ## Process


justaugustus (commented):
/hold
I think my original intent here was to allow Enhancements team members to do reviews in k/enhancements.

Do we want to try and do that for 1.21?

kikisdeliveryservice:
> /hold
> I think my original intent here was to allow Enhancements team members to do reviews in k/enhancements.
> 
> Do we want to try and do that for 1.21?

We can still review enhancements but the enhancements release team isn't really approving them.. I don't recall any of us ever having the power to do that and it... doesn't really make sense esp if we want to shift to sigs owning their enhancements more... as k/k members and milestone maintainers we are able to review etc... and I've not had a problem with sig-level approvers approving over my suggestions.

kikisdeliveryservice:
Wait @justaugustus just to clarify this is for the enhancements team in the release team  not the enhancements subproject for ref.

justaugustus:
> Wait @justaugustus just to clarify this is for the enhancements team in the release team not the enhancements subproject for ref.

Yep, I'm aware. :)

I see every subteam of the Release Team as an extension of some other group in the community.
In this case: Enhancements subteam --> Enhancements subproject (SIG Arch).

With that framing, I would conjecture that the subteam would want to be able to have the ability to review what's flowing into the subproject.
Note that I'm talking explicitly about things like Enhancements subproject policy and documentation, NOT the SIG-specific content in `keps/sig-**/**`.

justaugustus:
All of that to say, as an @kubernetes/enhancements subproject owner, I value the input of the contributors on the ground on the Enhancements subteam of the Release Team.

kikisdeliveryservice:
The problem is that enhancements is also keps/owners:
https://github.com/kubernetes/enhancements/blob/c26cc57548c1f9ca0abb51f57d2b1c6569b1baa6/keps/OWNERS

kikisdeliveryservice:
But wouldn't this communication also be leveraged via the github team that they are joining? (see https://git.k8s.io/org/config/kubernetes/sig-architecture/teams.yaml)

justaugustus:
> The problem is that enhancements is also keps/owners:
> https://github.com/kubernetes/enhancements/blob/c26cc57548c1f9ca0abb51f57d2b1c6569b1baa6/keps/OWNERS

Haha, that's a good route to notification fatigue (and probably why I didn't push on this in the past).
I took a note to revisit (maybe as part of a directory restructure e.g., `k/enhancements/process/OWNERS`), but no need to block on that here.

/hold cancel
/lgtm
/approve

PR #1412, opened 2021-01-12
Wrap up licensing subproject

dims:
The dependencies for kubernetes/kubernetes is in good shape, we can handle any items left as part of area/code-organization (sig-arch) if needed.

Signed-off-by: Davanum Srinivas <davanum@gmail.com>

#### What type of PR is this:

/kind cleanup


dims:
cc @justaugustus @swinslow @nikhita

hasheddan (approved):
/lgtm
/hold for those tagged

licensing/README.md
```diff
@@ -1,13 +0,0 @@
-# Licensing
```
nikhita, line 1: Should we move this doc to the sig-arch directory in k/community - https://github.com/kubernetes/community/tree/master/sig-architecture?

dims: probably better nearer to where we point folks to when they update a dependency?
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/vendor.md

saschagrunert (approved):
/lgtm

licensing/README.md
nikhita: > probably better nearer to where we point folks to when they update a dependency?

:+1:

nikhita: created https://github.com/kubernetes/community/pull/5400

PR #1443, opened 2021-01-31
kube-controller-manager is already distroless

dims:
https://github.com/kubernetes/kubernetes/issues/78737 is closed. fixed in https://github.com/kubernetes/kubernetes/pull/91329


hasheddan (approved):
Thanks @dims!

/lgtm

PR #1446, opened 2021-02-01
fix all 1.20/1.21 typos and grammer issues

pacoxu:
#### What type of PR is this:

/kind cleanup

move  https://github.com/kubernetes/kubernetes/pull/98047 here


hasheddan (commented):
@pacoxu thanks for the effort here! However, since the purpose of these documents are to demonstrate release note generation, I don't think we should be introducing the churn of making grammatical fixes. If these fixes need to be made it may make sense to edit the release note of the relevant PR in k/k. I'll leave final decision here to @wilsonehusin

pacoxu:
> @pacoxu thanks for the effort here! However, since the purpose of these documents are to demonstrate release note generation, I don't think we should be introducing the churn of making grammatical fixes. If these fixes need to be made it may make sense to edit the release note of the relevant PR in k/k. I'll leave final decision here to @wilsonehusin 

totally agree.

we need to re generate the release note than changing it frequently.

wilsonehusin:
@pacoxu Let me share a bit more details from my point of view since there has been lots of changes on CHANGELOG in the past week or so:

- **I'm interested in pre-release changes to make it to this repository** (currently, _pre-release_ is Kubernetes 1.21), because:
  - The whole document in k/k is generated from here.
  - Once `.0` is released, `alpha`, `beta`, `rc` release notes won't bring as much value as the actual release because there will be duplicates between all of them and `.0` is what matters the most, [see this commit for reference in 1.20.0](https://github.com/kubernetes/kubernetes/commit/00d6a12d02c0eb1ccd612752c195af3391aae776).
- Changes for versions prior to pre-release can go directly to k/k since they're not getting regenerated.
- Changes for [EOL versions](https://github.com/kubernetes/sig-release/blob/master/releases/patch-releases.md#non-active-branch-history) should not be made at all, I don't think it's a good idea to encourage activity on EOL releases.


So, having that said, for this PR, I think we should:
- [ ] Let versions other than 1.21 unchanged
- [ ] Create the changes for 1.21 in map format, [you can see the examples here](https://github.com/kubernetes/sig-release/tree/master/releases/release-1.21/release-notes/maps)

Is that helpful?

wilsonehusin:
@pacoxu related to the "map format" mentioned above, would you be interested to contribute in incorporating the grammar / spelling checker into our tooling? these release notes are parsed from PRs with our tooling which has the ability to override the values.

PR #1468, opened 2021-02-17
Migrate 1.21 release-notes-draft to release-notes/

wilsonehusin:
Signed-off-by: Wilson E. Husin <wilsonehusin@gmail.com>

#### What type of PR is this:

/kind cleanup

#### What this PR does / why we need it:

Following https://github.com/kubernetes/release/issues/1919
and will continue to be the behavior as of https://github.com/kubernetes/release/pull/1922

PR #1469, opened 2021-02-17
Add 1.20 release-notes-draft to release-notes directory

wilsonehusin:
Signed-off-by: Wilson E. Husin <whusin@vmware.com>

#### What type of PR is this:

/kind cleanup

Backporting #1468 as we need it for test completion (see https://github.com/kubernetes/release/pull/1922)

/cc @puerco @hasheddan


hasheddan (commented):
@wilsonehusin can we just modify the test to run against master branch?

wilsonehusin:
🤔 I was thinking about that too, but I worry that the behavior of the test would drift from the actual execution, since full release is taken from `release-x.y` branch, correct? or is that no longer the case because of fast-forwarding behavior that we started doing in 1.20?

hasheddan:
@wilsonehusin the changelog command allows for execution on an arbitrary repo, so we technically are not scoped to what is available in k/k, but I understand the sentiment. Another option would be to make the location of the generated notes configurable. Right now the [sig-release repo is hard-coded](https://github.com/kubernetes/release/blob/d71805402dab160d601bdbe32c5e9178dd82d29f/pkg/changelog/changelog.go#L406), but I could see the desire to point to another location in the future, potentially even locally. In general, these tests seem more like integration / e2e tests rather than the unit tests we are using them as. I would prefer that we don't clone all of k/k for example on each run when we could just construct a mock locally. Anyway -- what do you think about making the remote location configurable and using 1.20 for now?

wilsonehusin:
yeah I gave it more thought last night and I think we should probably just mock out the tests. I'll close this one for now.

PR #1498, opened 2021-03-17
Remove "next patch release is X" lines

jberkus:
The "Next Patch Release is #.#" line in each major release frequently does not get updated when the other patch information on this page gets updated.  In the current page, it's behind for all three supported releases.  Since that information is entirely redundant, it seems simpler to just delete that line so people don't have to remember to update it.

### DIFF
--- a/releases/patch-releases.md
+++ b/releases/patch-releases.md
@@ -57,8 +57,6 @@ releases may also occur in between these.
 
 ### 1.20
 
-Next patch release is **1.20.3**
-
 End of Life for **1.20** is **2021-12-30**
 
 | PATCH RELEASE | CHERRY PICK DEADLINE | TARGET DATE |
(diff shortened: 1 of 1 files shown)


jberkus:
attn: @justaugustus

releases/patch-releases.md
```diff
@@ -57,8 +57,6 @@ releases may also occur in between these.
 
 ### 1.20
 
-Next patch release is **1.20.3**
```
hasheddan, line 60: This seems like a good change to me given the current workflow, but I think we should look into automatically updating this page in the future. Seems like it could be a pretty easy modification of some existing release tooling 👍

hasheddan (approved):
/lgtm

releases/patch-releases.md
puerco: I think @cpanato was planning something like that. Let's chat the next time we get together.

cpanato: i was planning/working in the notification system, but we can add a flow to update this page automagically with a prow job. But we will need some manual input to trigger this, maybe we update the schedule yaml file and when we merge that it triggers a prwo job to update the page?

PR #1515, opened 2021-04-09
Add 1.21 to the patch releases schedule

xmudrii:
#### What type of PR is this:

/kind documentation

The Kubernetes 1.21 release went out yesterday, so this PR adds 1.21 to the patch releases schedule.

Ref: https://kubernetes.slack.com/archives/C2C40FMNF/p1617961638360800

None

cc @kubernetes/release-engineering @kubernetes/release-team-leads

### DIFF
--- a/releases/patch-releases.md
+++ b/releases/patch-releases.md
@@ -54,6 +54,14 @@ releases may also occur in between these.
 
 ## Detailed Release History for Active Branches
 
+### 1.21
+
+End of Life for **1.21** is **2022-04-30**
+
+| PATCH RELEASE | CHERRY PICK DEADLINE | TARGET DATE |
+|--- |--- |--- |
+| 1.21.1        | 2021-05-07           | 2021-05-12  |
+
 ### 1.20
 
 End of Life for **1.20** is **2021-12-30**
--- a/releases/schedule.yaml
+++ b/releases/schedule.yaml
@@ -1,4 +1,10 @@
 schedules:
+- release: 1.21
+  next: 1.21.1
+  cherryPickDeadline: 2021-05-07
+  targetDate: 2021-05-12
+  endOfLifeDate: 2022-04-30
+  previousPatches:
 - release: 1.20
   next: 1.20.6
   cherryPickDeadline: 2021-04-09


releases/patch-releases.md
```diff
@@ -54,6 +54,14 @@ releases may also occur in between these.
 
 ## Detailed Release History for Active Branches
 
+### 1.21
+
+End of Life for **1.21** is **2022-04-30**
```
xmudrii, line 59: I'm not sure is this the correct date, but I assumed one year from now, so tentatively it should be by end of April 2022.

releases/patch-releases.md
```diff
@@ -54,6 +54,14 @@ releases may also occur in between these.
...
+
+| PATCH RELEASE | CHERRY PICK DEADLINE | TARGET DATE |
+|--- |--- |--- |
+| 1.21.1        | 2021-05-07           | 2021-05-12  |
```
xmudrii, line 63: I assume that we will not cut 1.21.1 next week, since 1.21.0 went out yesterday. Is that correct or not?

cc @kubernetes/release-managers

releases/patch-releases.md
saschagrunert: You're right.

Generally we may omit the actual day from the EOL date. I think this way we can simply state that the last patch release  is the one of April 2022. :thinking:

releases/patch-releases.md
saschagrunert: I agree.

cpanato: agree, we will release just 1.18/1.19/1.20

cpanato (approved):
/lgtm

hasheddan (commented):
/hold until @saschagrunert comments are addressed

releases/patch-releases.md
xmudrii: @saschagrunert @hasheddan I thought about that as well, but I was not sure what to put in `schedule.yaml`. That file is supposed to be machine-readable, and so far we've only put dates in it. I thought about omitting day (e.g. just `2022-04`) or putting some string description, but I'm worried it could break users that depend on that file (if there are any).

I wanted to keep the consistency between `patch-releases.md` and `schedule.yaml`, so I decided to put some tentative date as we did before. If you have any idea how to workaround this, please let me know.

saschagrunert: Let's keep them for now and revisit this discussion later on.

saschagrunert (approved):
/hold cancel

saschagrunert:
@xmudrii I think we should communicate updates to the patch release schedule to the mailing list. What do you think?

xmudrii:
@saschagrunert I think that this makes sense. Should we communicate this change to the mailing list as well, or should we only start doing that in the future?

saschagrunert:
> @saschagrunert I think that this makes sense. Should we communicate this change to the mailing list as well, or should we only start doing that in the future?

I think you can start by proposing this changes while mentioning that we will also do it in the future :)

PR #1525, opened 2021-04-16
Add Release Team Diagram

JamesLaverack:
#### What type of PR is this:

/kind documentation

Adds an SVG diagram that shows the roles in the release team. This is to aid the text it sits alongside in describing the team's structure.


JamesLaverack:
I realise that I didn't consider dark mode users. It looks as intended in light mode:
Screenshot 2021-04-16 at 22 12 12
but in dark mode the "fade" effect does not work and the text without a background is almost invisible:
image

annajung:
I think we should get rid of "typically 3-4 shadows per area" since there have been talks of increasing the number of shadows and leaving it flexible for each sub Leads due to the increasing amount of enhancements past few releases. 

I think the dark mode looks okay especially if we get rid of the text then it wouldn't be a problem that it doesn't show up :)

JamesLaverack:
> I think we should get rid of "typically 3-4 shadows per area" since there have been talks of increasing the number of shadows and leaving it flexible for each sub Leads due to the increasing amount of enhancements past few releases.

:+1: I've removed it.

I've updated it to use opacity instead of distinct 'lighter' colours anyway, so that it'll work on dark backgrounds much better.

Screenshot 2021-04-21 at 01 05 01
Screenshot 2021-04-21 at 01 05 51

PI-Victor:
was about to suggest a gray color, but i agree with the suggestion to remove it from there.
thanks for the PR, i guess we can merge it?
/lgtm

palnabarun:
/lgtm
/approve

Thanks @JamesLaverack 🎉

PR #1532, opened 2021-04-22
patches: Add maintenance mode dates, updated EOLs, and final v1.18 release

justaugustus:
#### What type of PR is this:

/kind cleanup documentation feature

- Restore release history table for v1.18 series
- Add a final maintenance release (v1.18.19) to the v1.18 series
- Extend EOL dates to include a maintenance window
- Add maintenance mode dates for active release series
- Add copy around maintenance mode and EOL expectations

/assign @hasheddan @saschagrunert @xmudrii 
cc: @kubernetes/sig-release-leads @kubernetes/release-managers 

ref:

- https://kubernetes.slack.com/archives/CJH2GBF7Y/p1619041703010600
- https://kubernetes.slack.com/archives/C2C40FMNF/p1618918209102100
- https://kubernetes.slack.com/archives/C2C40FMNF/p1619016765130500
- https://kubernetes.slack.com/archives/CJH2GBF7Y/p1618852608477700


releases/patch-releases.md
```diff
@@ -93,14 +127,40 @@ End of Life for **1.19** is **2021-09-30**
...
 
+### 1.18
+
+**1.18** enters maintenance mode on **2021-04-28**
+
+End of Life for **1.18** is **2021-04-30**
+
+| PATCH RELEASE | CHERRY PICK DEADLINE | TARGET DATE |
+|--- |--- |--- |
+| 1.18.19       | 2021-04-26           | 2021-04-28  |
```
hasheddan, line 138: I would like to propose final patch release for 1.18 on 5/7/21 with the rest of the active branch patches. This is simpler for folks trying to get cherry picks backported to all branches, and also provides more time given that we are a bit late communicating this date.

puerco:
Much clearer, thanks 😊
/lgtm

releases/patch-releases.md
justaugustus: @hasheddan -- I'd like to minimize how much longer we keep the 1.18 branch around, as it's running on go1.13.15 (which is out of support upstream).

The thought process around the 28th was although it deviates from the patch release schedule month-to-month, it allows us to give a very consistent response about when a release in entering maintenance mode or going EOL; it's _always_ the 28th of the designated month.

A consideration for a follow-up 👇🏾 
Are we okay with 5/7 as a patch release date next month? That falls within the week of KubeCon.

hasheddan: That sounds like a good policy (having final patch be on the 28th of the EOL month), but I would encourage us to consider enacting that post 1.18 as we are already close to the deadline and some folks have made the assumption (reasonably) that if we were to have one more patch it would be aligned with the other branches. We should also consider any future confusion this could introduce with folks opening CPs to branches that have different CP deadlines when the oldest active branch is reaching EOL.

justaugustus: > some folks have made the assumption (reasonably) that if we were to have one more patch it would be aligned with the other branches

@hasheddan -- Great point. Updated!

> Are we okay with 5/7 as a patch release date next month? That falls within the week of KubeCon.

Don't mind me. Looking again, I see that this is the cherry pick deadline, with the patches to happen the week after (which should be fine).

cpanato (approved):
thanks for update and make sense to me

/lgtm

saschagrunert (approved):
/lgtm

Let's wrap-up those changes in a mail, too.

xmudrii:
Should we also update [`schedule.yaml`](https://github.com/kubernetes/sig-release/blob/master/releases/schedule.yaml) or should we do it in a follow-up?

releases/patch-releases.md
```diff
@@ -39,6 +39,34 @@ of the actual release.
...
+
+- [Release Managers][release-managers] will cut a release
+- The patch release series will enter maintenance mode
+
+During the two-month maintenance mode period, Release Managers may cut
+additional maintenance releases to resolve:
+
+- CVEs (under the advisement of the Product Security Committee)
+- dependency issues (including base image updates)
+- critical core component issues
```
xmudrii, line 61: Are we going to allow other changes in the maintenance mode period?

I guess that one option is that we might allow them if we are fixing one of those three issues specified here, but otherwise, we wouldn't cut a release only because of other changes. Another option is that we don't allow other changes non-related to those at all.

Should we make that clear in the document?

xmudrii (approved):
Other than the comment I left, this looks great to me. Thank you for taking care of this!

/lgtm

hasheddan (approved):
/lgtm

justaugustus:
Note to the community to pre-announce May's patch releases and extending EOLs with a maintenance mode: https://groups.google.com/g/kubernetes-announce/c/Vixft2lPyaA

PR #1535, opened 2021-04-23
Fix typo in 1.20 maintenance mode date

reingart:
#### What type of PR is this:

/kind documentation

This fixes the Kubernetes 1.20 maintenance mode date that seems to be off by 1 year

It seems that 1.20 has a typo (12 months after release is Dec 2021)
Also EOL is Feb 2022, so it cannot be earlier than maintenance deadline.

### DIFF
--- a/releases/patch-releases.md
+++ b/releases/patch-releases.md
@@ -93,7 +93,7 @@ End of Life for **1.21** is **2022-06-28**
 
 ### 1.20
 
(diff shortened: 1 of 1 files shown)


reingart:
I have signed the Contributor License Agreement (CLA) a few minutes ago, and received the confirmation from HelloSign

justaugustus:
Thanks for catching this, @reingart!
/lgtm
/approve

PR #1579, opened 2021-05-21
Clarify enhancements lead time requirements

JamesLaverack:
#### What type of PR is this:

/kind documentation

#### What this PR does / why we need it:

The time requirements documented for Enhancements Lead are... low. During my time as Enhancements Lead for Kubernetes v1.22, I logged the time I spent on upstream work and came with the following:

- Before the release (w/c 12th April 2021): 3 hours
- Week 0 (w/c 19th April 2021): 6.5 hours
- Week 1 (w/c 26th April 2021): 15 hours
- Week 2 (w/c 3rd May 2021): 10 hours
- Week 3 (w/c 10th May 2021, enhancements freeze week): 25.5 hours
- Week 4 (w/c 17th May 2021, to date): 3 hours

Given this data, and [discussions with former Enhancements Leads](https://kubernetes.slack.com/archives/C2C40FMNF/p1621551220054200), I'm increasing the documented time requirements. I'm also clarifying the requirements for shadows.

PR #1581, opened 2021-05-21
Clarify enhancements lead time requirements

JamesLaverack:
#### What type of PR is this:

/kind documentation

The time requirements documented for Enhancements Lead are... low. During my time as Enhancements Lead for Kubernetes v1.22, I logged the time I spent on upstream work and came with the following:

- Before the release (w/c 12th April 2021): 3 hours
- Week 0 (w/c 19th April 2021): 6.5 hours
- Week 1 (w/c 26th April 2021): 15 hours
- Week 2 (w/c 3rd May 2021): 10 hours
- Week 3 (w/c 10th May 2021, enhancements freeze week): 25.5 hours
- Week 4 (w/c 17th May 2021, to date): 3 hours

Given this data, and [discussions with former Enhancements Leads](https://kubernetes.slack.com/archives/C2C40FMNF/p1621551220054200), I'm increasing the documented time requirements. I'm also clarifying the requirements for shadows.

This was originally submitted as https://github.com/kubernetes/sig-release/pull/1579 but was accidentally merged early. This was reverted in https://github.com/kubernetes/sig-release/pull/1580 so it could be re-proposed in this PR.

### DIFF
--- a/release-team/role-handbooks/enhancements/README.md
+++ b/release-team/role-handbooks/enhancements/README.md
@@ -43,10 +43,17 @@ Helpful characteristics of an Enhancements Lead include:
 - product / project / program management experience
 - release management experience
 
-Approximate Time Commitments
-- Beginning of the cycle through enhancement freeze: 6-10 hours a week fluctuating based on how many SIG meetings need to be attended
-- Enhancement Freeze through Code Freeze: 4-7 hours a week
-- Code Freeze through Release Day: 1-4 hours a week
+### Time Commitments
+
+Enhancement is one of the most time-intensive areas of the release team, and especially so during the early parts of the release. An Enhancements Lead can expect to spend:
+- Beginning of the cycle through enhancement freeze: ~8–15 hours a week
+- Week of enhancements freeze: 20+ hours
+- Enhancement Freeze through Code Freeze: ~4–7 hours a week
+- Code Freeze through Release Day: ~1–4 hours a week
+
+Note that Enahncements Lead in particular will need to do work **during the week** during the early release, and will need to be available at least daily.
+
+Enhancements shadows have much lower time committments, but can still expect to spend 5+ hours a week during the early release, and ~1–4 hours a week after enhancements freeze. Unlike Enhancements Lead, shadows can expect to do their work largely on weekends if they desire.
 
 ## Mentoring Shadows
 
--- a/release-team/role-handbooks/enhancements/README.md
+++ b/release-team/role-handbooks/enhancements/README.md
@@ -123,7 +123,7 @@ It is important that this process be followed and documentation remain up-to-dat
 - Work with the Release Lead to introduce yourself, talk about release information, and relay information about opting into the release with SIG Leads.
 - Weekly Release meetings require updates of current status. Use the `Feature stats` tab to update the release team on counts of enhancements in good and bad progress.
 
-#### Collection Monitoring and Triage (Week 2)
+#### Collection Monitoring and Triage (Week 1-2)
 
 - Monitor the `KEP Collection` tab regularly for new additions as SIGs opt-in KEPs into the release.
 - Move data from `KEP Collection` into the `data` tab to be tracked officially for the release.
--- a/release-team/role-handbooks/enhancements/README.md
+++ b/release-team/role-handbooks/enhancements/README.md
@@ -53,7 +53,7 @@ Enhancement is one of the most time-intensive areas of the release team, and esp
 
 Note that Enahncements Lead in particular will need to do work **during the week** during the early release, and will need to be available at least daily.
 
-Enhancements shadows have much lower time committments, but can still expect to spend 5+ hours a week during the early release, and ~1–4 hours a week after enhancements freeze. Unlike Enhancements Lead, shadows can expect to do their work largely on weekends if they desire.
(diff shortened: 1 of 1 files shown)


release-team/role-handbooks/enhancements/README.md
```diff
@@ -43,10 +43,17 @@ Helpful characteristics of an Enhancements Lead include:
...
+
+Enhancement is one of the most time-intensive areas of the release team, and especially so during the early parts of the release. An Enhancements Lead can expect to spend:
+- Beginning of the cycle through enhancement freeze: ~8–15 hours a week
+- Week of enhancements freeze: 20+ hours
+- Enhancement Freeze through Code Freeze: ~4–7 hours a week
+- Code Freeze through Release Day: ~1–4 hours a week
+
+Note that Enahncements Lead in particular will need to do work **during the week** during the early release, and will need to be available at least daily.
+
+Enhancements shadows have much lower time committments, but can still expect to spend 5+ hours a week during the early release, and ~1–4 hours a week after enhancements freeze. Unlike Enhancements Lead, shadows can expect to do their work largely on weekends if they desire.
```
annajung, line 56: This is where we can really set the shadow expectation for the future cycles. IMO, shadows should plan to spend **more** than min 5 hours/week especially at the beginning of the cycle, especially a week leading up to enhancements freeze. That will help grow a better enhancements team and a better successor for future cycles as they will have more time to get embedded in the role.

IMO, we should expect 10+ hours (maybe even 15+) for shadows at the beginning of the cycle. Leads are responsible for double-checking shadows work which requires more commitment, but shadows should also be taking ownership of the assigned tasks related to the enhancements. We should clearly put that expectation here. This can also help distribute the load from Lead, allowing more flexibility of time commitment for the Lead. I just think 20+ hours for Lead and 5+ hours for shadows is not a good balance.

Maybe current/previous enhancements shadows can suggest the hours for shadows? especially @jrsapi (1.21 and 1.22 enhancements shadow), I think your feedback would be really valuable here :)

release-team/role-handbooks/enhancements/README.md
```diff
@@ -116,7 +123,7 @@ It is important that this process be followed and documentation remain up-to-dat
 - Work with the Release Lead to introduce yourself, talk about release information, and relay information about opting into the release with SIG Leads.
 - Weekly Release meetings require updates of current status. Use the `Feature stats` tab to update the release team on counts of enhancements in good and bad progress.
 
-#### Collection Monitoring and Triage (Week 2-3)
+#### Collection Monitoring and Triage (Week 2)
```
annajung, line 126:
```suggestion
#### Collection Monitoring and Triage (Week 1-2)
```

release-team/role-handbooks/enhancements/README.md
annajung: also cc the rest of the current shadows @salaxander @reylejano @gracenng @supriya-premkumar
Feedback on the time commitment for shadows would be appreciated! thanks!

JamesLaverack: Thanks for the good ideas @annajung. I'll survey a few recent Enhancements shadows and try to see if there's anyone who'd like to share their viewpoints. I agree we could do more to involve shadows more, but of course that requires more time commitment.

gracenng: I agree with Anna on balancing out the time commitment between lead and shadows.

Personally, the two weeks leading up to Enhancement Freeze I was putting in 10+ hours a week and that aligned with what I expected. James did an amazing job keeping us in shape by picking up enhancements that pops up here and there, so he had almost 2x the issues we do. Certainly this amount should be redistributed in the future and in 1.22 Code Freeze 😄

JamesLaverack: I don't have data, but I think I spent about the same as an enhancements shadow in v1.21. What kind of time do we think is appropriate for the shadows? Maybe about half of what the lead has as an expectation? Or move them closer together even. We could give general time commitments for all members, and then say (the lead can expect to spend +X% more).

jrsapi: I've had the experience from a previous cycle but found myself spending about 10 hours in weeks 1 and 2. Then when we got close to freeze it was about 15 hours. We should definitely raise from five.

annajung: 10-15 hours a week or two weeks before the enhancements freeze might be more realistic from what @gracenng and @jrsapi shared. By updating the doc to reflect the correct amount of time will ensure correct expectations for shadows.

Was there any increase in hours spent for code freeze? That might be also good to insert.

I don't think this is necessarily true `Enhancements shadows have much lower time committments`, rather, their work is mostly loaded in the beginning.

JamesLaverack: I've reworked this paragraph to try to reflect the outcome of this discussion. Could folk give it another look? cc @annajung @reylejano @salaxander @jrsapi @gracenng

palnabarun:
The changes look good! 👍🏽

Thanks for tracking and calculating the time commitments!

/lgtm
/approve

palnabarun:
@JamesLaverack -- please feel free to remove the hold whenever you feel so!

JamesLaverack:
Ack. I'll give it a last look this evening. :)

PR #1584, opened 2021-05-24
Update release notes draft to version v1.22.0-alpha.2

Damans227:
**What type of PR is this?**
/kind documentation

**What this PR does / why we need it**:
This PR updates the Release Notes Draft to k/k v1.22.0-alpha.2

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:
This is an automated PR generated from `krel The Kubernetes Release Toolbox`

PR #1593, opened 2021-06-11
Quick fix for the yaml file parsing

cici37:
#### What type of PR is this:
/kind bug

This is a bug introduced from https://github.com/kubernetes/sig-release/pull/1584, and causing issue when generating release note for v1.22-alpha3: https://github.com/kubernetes/release/issues/2114

https://github.com/kubernetes/release/issues/2114

### DIFF
--- a/releases/release-1.22/release-notes/maps/pr-101496-map.yaml
+++ b/releases/release-1.22/release-notes/maps/pr-101496-map.yaml
@@ -1,4 +1,4 @@
 pr: 101496
 releasenote:
-  text: `podAffinity`, `NamespaceSelector` and the associated `CrossNamespaceAffinity`
-    quota scope features graduate to Beta and they are now enabled by default.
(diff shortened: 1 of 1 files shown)


releases/release-1.22/release-notes/maps/pr-101496-map.yaml
```diff
@@ -1,4 +1,4 @@
 pr: 101496
 releasenote:
-  text: `podAffinity`, `NamespaceSelector` and the associated `CrossNamespaceAffinity`
-    quota scope features graduate to Beta and they are now enabled by default.
+  text: "`podAffinity`, `NamespaceSelector` and the associated `CrossNamespaceAffinity`
```
pmmalinov01, line 3: Can you put a word before  `podAffinity`, something like The or A ?

cici37: Updated. Would you mind to review again? Thank you

PR #1613, opened 2021-07-06
Remove SIG PM from README files

whytoe:
Removes SIG PM references from the various README.md files as SIG PM has been retired

/kind documentation
/kind cleanup

/hold
(for reviews)


puerco:
Thank you @whytoe !
/lgtm
/approve
/hold cancel

PR #1642, opened 2021-07-29
ci-signal: use k8s-release-dev for CI builds

spiffxp:
#### What type of PR is this:

/kind cleanup
/kind documentation

kubernetes-release-dev is deprecated and has no new builds as of v1.23

n/a

Related:
- Part of: https://github.com/kubernetes/k8s.io/issues/2318
- Migration of similar content in kubernetes/community: https://github.com/kubernetes/community/pull/5893

### DIFF
--- a/release-team/role-handbooks/ci-signal/upgrade-job-versions.md
+++ b/release-team/role-handbooks/ci-signal/upgrade-job-versions.md
@@ -36,10 +36,10 @@ aka trust the CI
 - look for the `--extract=` flags
 - the version there corresponds to files in GCS buckets
 
-ci/foo versions live in kubernetes-release-dev:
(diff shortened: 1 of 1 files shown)


justaugustus (approved):
/lgtm
/approve

@spiffxp -- as a follow-up, should we prefer `curl` in docs over `gsutil`?

spiffxp:
I mean there's a part of me that wonders if this shouldn't just link to https://github.com/kubernetes/community/blob/master/contributors/devel/sig-release/kubernetes-versions.md instead of having similar content

But I defer that to folks here, I'm trying to keep my attention focused elsewhere vs sprawling into doc rewrites today

PR #1671, opened 2021-08-19
docs: add a place where retro link should be shared

tpepper:
The release retrospectives are linked from the header of the
Kubernetes Community Meeting Topics (who knew?!).  These hadn't
been updated since the 1.18 release, which is a sign there isn't a
clear owner for doing that updating.  It's really useful to have
early and broad visibility for retrospectives in order to capture
folks input during the cycle, so I'm proposing a specific owner on
it.


cpanato (approved):
thanks Tim

/lgtm

PR #1677, opened 2021-08-23
Update v1.23.0-alpha.1 cut in release calendar

puerco:
#### What type of PR is this:
/kind documentation

The release cut for the alpha 1 of Kubernetes 1.23 already happened
this commit modifies the calendar to correct the date.

Signed-off-by: Adolfo García Veytia (Puerco) <adolfo.garcia@uservers.net>

None

/assign @reylejano

### DIFF
--- a/releases/release-1.23/README.md
+++ b/releases/release-1.23/README.md
@@ -52,7 +52,6 @@ The 1.23 release cycle is proposed as follows:
 |---|---|-------|---|---|
 | Start of Release Cycle | Lead | Mon August 23, 2021 | week 1 | [master-blocking] |
 | Start Enhancements Tracking | Enhancements Lead | Mon August 23, 2021| week 1 | |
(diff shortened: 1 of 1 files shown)


puerco:
I moved the alpha.1 line to the first spot to get it in chronological order, but something doesn't feel quite right not having the release start date as the first entry. WDYT @reylejano  ?

reylejano:
Since it's before the release cycle, it might be better to delete the entry for alpha.1 since it was cut before the release cycle.
Using 1.19 as an example where it starts with alpha.2
https://github.com/kubernetes/sig-release/blob/master/releases/release-1.19/README.md

/area release-team
/sig release
/assign @justaugustus @saschagrunert @jeremyrickard @cpanato 
/cc @salaxander @encodeflush @voigt @jlbutler @cici37 @karenhchu

puerco:
Sounds good to me. I've removed the `alpha.1` line to encompass only events within the cycle. PTAL

PR #1813, opened 2022-01-12
Update OWNERS_ALIASES for 1.24

JamesLaverack:
#### What type of PR is this:

/kind cleanup
/priority important-soon

As per discussion on Slack: https://kubernetes.slack.com/archives/C2C40FMNF/p1642000175084300

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -24,55 +24,52 @@ aliases:
     - thejoycekung # Release Manager Associate
     - wilsonehusin # Release Manager Associate
   release-team:
-    - jeremyrickard # subproject owner / 1.20 RT Lead
-    - justaugustus # subproject owner
-    - onlydole # subproject owner / 1.19 RT Lead
-    - palnabarun # subproject owner / 1.21 RT Lead
-    - reylejano # 1.23 RT Lead
-    - savitharaghunathan # subproject owner / 1.22 RT Lead
+    - justaugustus # SIG Release Chair
+    - saschagrunert # SIG Release Chair
+    - cpanato # SIG Release Technical Lead
+    - jeremyrickard # SIG Release Technical Lead
+    - puerco # SIG Release Technical Lead
+    - palnabarun # Kubernetes 1.21 Release Lead
+    - savitharaghunathan # Kubernetes 1.22 Release Lead
+    - reylejano # Kubernetes 1.23 Release Lead
+    - jameslaverack # Kubernetes 1.24 Release Lead
   build-admins:
     - amwat
     - BenTheElder
     - MushuEE
     - spiffxp
   release-team-lead-role:
-    - onlydole # 1.19 RT Lead / 1.21 Emeritus Advisor
-    - jeremyrickard # 1.20 RT Lead
-    - palnabarun # 1.21 RT Lead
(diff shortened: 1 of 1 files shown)


OWNERS_ALIASES
```diff
@@ -24,55 +24,52 @@ aliases:
...
-    - justaugustus # subproject owner
-    - onlydole # subproject owner / 1.19 RT Lead
-    - palnabarun # subproject owner / 1.21 RT Lead
-    - reylejano # 1.23 RT Lead
-    - savitharaghunathan # subproject owner / 1.22 RT Lead
+    - justaugustus # SIG Release Chair
+    - saschagrunert # SIG Release Chair
+    - cpanato # SIG Release Technical Lead
+    - jeremyrickard # SIG Release Technical Lead
+    - puerco # SIG Release Technical Lead
```
justaugustus, line 31: Prefer "subproject owner" for the comment for Chairs/TLs.
Reasoning: Subproject ownership survives leadership changes unless the contributor opts to step down

OWNERS_ALIASES
```diff
@@ -24,55 +24,52 @@ aliases:
...
+    - palnabarun # Kubernetes 1.21 Release Lead
+    - savitharaghunathan # Kubernetes 1.22 Release Lead
+    - reylejano # Kubernetes 1.23 Release Lead
+    - jameslaverack # Kubernetes 1.24 Release Lead
```
justaugustus, line 35: Prefer "x.yy Release Team Lead" for these comments.

justaugustus (changes requested):
@JamesLaverack -- tiny nits. Let's also alpha-sort these lists.

justaugustus:
> Let's also alpha-sort these lists.

(as a separate commit, so the diff is easy to read)

OWNERS_ALIASES
JamesLaverack: Does that mean I should *not* add @cpanato and @puerco? Does it mean I shouldn't remove @savitharaghunathan, @palnabarun, and @onlydole that were previously listed as "subproject owners"?

justaugustus: > Does that mean I should _not_ add @cpanato and @puerco?

Let's keep all leads in there.

> Does it mean I shouldn't remove @savitharaghunathan, @palnabarun, and @onlydole that were previously listed as "subproject owners"?

Leave them in. I've been taking "Release Team lead" to also mean "subproject owner for the lifecycle of the release they've led" (which is something we should probably document, if it isn't already).

The reasoning behind this is to ensure some continuity in reviews.
The active release team lead will primarily be engrossed with operationalizing the release, while previous RT leads have the opportunity to provide longer tail reviews.

JamesLaverack:
I've alpha sorted them. I don't think my editor has a automatic option so I did so manually and likely made a mistake somewhere, so do check. 😅

justaugustus (approved):

OWNERS_ALIASES
justaugustus: > I'm not sure I *fully* understand the distinction between a subproject owner, a release team lead, and a SIG Release Chair/TL. But I've updated it accordingly.

@jameslaverack -- can you file an issue for this? If it's not widely understood, it's worth clarifying.

PR #1858, opened 2022-02-23
fix yaml errors in mapping files for 1.24

csantanapr:
While running `krel release-notes` having yaml files in mappings will cause the tool to fail

The error would like this
```
FATA creating website PR: generating release notes in JSON format: gathering release notes: listing release notes: checking if a map exists for PR 108093: while reading release notes maps: while parsing note map in /var/folders/gl/l4j1f7bj0hlb83kk3_9wx4jm0000gn/T/k8s-2399869277/releases/release-1.24/release-notes/maps/pr-105964-map.yaml: decoding note map: yaml: line 3: found character that cannot start any token
```


puerco:
Thanks for fixing this!
/lgtm
/approve

PR #1885, opened 2022-04-14
Add Kubernetes 1.25 release draft outline 

cici37:
#### What type of PR is this:
/kind documentation

Add the initial outline for the Kubernetes 1.25 release cycle.
Raise it early so we could kickoff the discussion on the dates.

### DIFF
--- a/releases/release-1.25/README.md
+++ b/releases/release-1.25/README.md
@@ -1,0 +1,118 @@
+---
+title: "Kubernetes 1.25 Release Information"
+weight: 98
+slug: "release"
+aliases: [ "/release" ]
+description: |
+  Information regarding the current release cycle including important dates,
+  Release Team contact information, tracking spreadsheets and more!
+---
+
+# Kubernetes 1.25
+
+#### Links
+
+* [This document](https://git.k8s.io/sig-release/releases/release-1.25/README.md)
+* [Release Team](https://github.com/kubernetes/sig-release/blob/master/releases/release-1.25/release-team.md)
+* [Meeting Minutes](http://bit.ly/k8s125-releasemtg) (members of [release-team@] receive meeting invites)
+* [v1.25 Release Calendar][k8s125-calendar]
+* Contact: [#sig-release] on slack, [release-team](mailto://release-team@kubernetes.io) on e-mail
+* [Internal Contact Info] (accessible only to members of [release-team@])
+
+#### Tracking docs
+
+* [Enhancements Tracking Sheet](https://bit.ly/k8s125-enhancements)
+* [Feature blog Tracking Sheet](http://bit.ly/k8s125-feature-blog)
+* [Bug Triage Tracking Sheet](https://bit.ly/k8s125-bug-triage-tracking)
+* [CI Signal Project Board & Issue Tracking](https://github.com/orgs/kubernetes/projects/68)
+* [Retrospective Document][Retrospective Document]
+* [kubernetes/sig-release v1.25 milestone](https://github.com/kubernetes/kubernetes/milestone/58)
+
+#### Guides
+
+* [Targeting Issues and PRs to This Milestone](https://git.k8s.io/community/contributors/devel/sig-release/release.md)
+* [Triaging and Escalating Test Failures](https://git.k8s.io/community/contributors/devel/sig-testing/testing.md#troubleshooting-a-failure)
+
+## Summary
+
+The 1.25 release cycle is as follows:
+
+- **Monday 9th May 2022**: Week 1 — Release cycle begins
+- **Thursday 26th May 2022**: Week 3 — [Production Readiness Soft Freeze](https://groups.google.com/g/kubernetes-sig-architecture/c/a6_y81N49aQ)
+- **02:00 UTC Friday 3rd June 2022** / **18:00 PST Thursday 2nd June 2022** : Week 4 — [Enhancements Freeze](../release_phases.md#enhancements-freeze)
+- **Wednesday 20th July 2022**: Week 11 — [Release Retrospective][Retrospective Document] part 1
+- **01:00 UTC Wednesday 27th July 2022** / **02:00 BST Wednesday 27th July 2022** / **18:00 PDT Tuesday 26th July 2022**: Week 12 — [Code Freeze](../release_phases.md#code-freeze)
+- **01:00 UTC Wednesday 3rd August 2022** / **02:00 BST Wednesday 3rd August 2022** / **18:00 PDT Tuesday 2nd August 2022**: Week 13 — [Test Freeze](../release_phases.md#test-freeze)
+- **Tuesday 9th August 2022**: Week 14 — Docs must be completed and reviewed
+- **Tuesday 16th August 2022**: Week 15 — Kubernetes v1.25.0 released
+- **TBC**: Week 16 — [Release Retrospective][Retrospective Document] part 2
+
+## Timeline
+
+| **What** | **Who** | **When** | **Week** | **CI Signal** |
+|---|---|-------|---|---|
+| Start of Release Cycle | Lead | Monday 9th May 2022 | week 1 | [master-blocking] |
+| Start Enhancements Tracking | Enhancements Lead | Monday 9th May 2022 | week 1 | |
+| Schedule finalized | Lead | Friday 13th May 2022 | week 1 | |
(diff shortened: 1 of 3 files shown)


reylejano:
@kubernetes/sig-release-leads , thoughts on the v1.25 release cycle starting on May 2 and scheduled release on August 16?

If May 2 starts week 1 of the release cycle, August 16 is in week 16 of the release cycle (with week week 3 being KubeCon Europe). (if I counted correctly :) )

The original schedule according to the [release cadence KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-release/2572-release-cadence), is to start v1.25 release cycle on April 26 and release on August 9.

We can push back the scheduled v1.25 release to August 9 and keep a May 2 start date for a 15 week cycle.

Thoughts?

JamesLaverack:
> We can push back the scheduled v1.25 release to August 9 and keep a May 2 start date for a 15 week cycle.

Would that put too much squeeze on enhancements freeze? I'd _speculate_ that most reviewers for KEPs, and the production readiness review team, are going to be busy on the lead-up to and during KubeCon. So realisticly enhancement owners have only two weeks after KubeCon EU to get their KEPs reviewed, and only one week to get it in for the PRR soft-freeze.

Unless of course you push enhancements freeze back by one week, which then reduces the "implementation period" between enhancements freeze and code freeze. 🤔

jeremyrickard:
> Would that put too much squeeze on enhancements freeze? I'd speculate that most reviewers for KEPs, and the production readiness review team, are going to be busy on the lead-up to and during KubeCon. So realisticly enhancement owners have only two weeks after KubeCon EU to get their KEPs reviewed, and only one week to get it in for the PRR soft-freeze.

KEP authoring/reviewing could be occurring now as well, so it's not like there is a "start" date for reviewing and things are condensed into just those two weeks. If we communicate now and say hey folks, here are the dates please work accordingly I don't think it's that big of an ask.

saschagrunert:
> thoughts on the v1.25 release cycle starting on May 2 and scheduled release on August 16?

According to the updated schedule for 1.24 we release on May 3rd, I'd like to avoid overlapping release cycles.

cici37:
Thanks for the review! I currently set the 1.25 release starts `Monday 9th May 2022` with the concern of not overlapping 1.24 release and give release team a break before starting 1.25. Please feel free to share thoughts on 1.25 release schedule. Thank you ^_^ @kubernetes/release-team-leads

releases/release-1.25/README.md
```diff
@@ -0,0 +1,118 @@
...
+#### Guides
+
+* [Targeting Issues and PRs to This Milestone](https://git.k8s.io/community/contributors/devel/sig-release/release.md)
+* [Triaging and Escalating Test Failures](https://git.k8s.io/community/contributors/devel/sig-testing/testing.md#troubleshooting-a-failure)
+
+## Summary
+
+The 1.25 release cycle is as follows:
+
+- **Monday 9th May 2022**: Week 1 — Release cycle begins
```
saschagrunert, line 40: In the cadence KEP we proposed to have a break of 2 weeks between releases: https://github.com/kubernetes/enhancements/tree/master/keps/sig-release/2572-release-cadence#proposal

> This KEP proposes a transition to a 3-releases-per-calendar-year cadence, beginning with the Kubernetes 1.22 Release. This would result in a 15 week release cycle, with 2 weeks between release cycles. 

My question would be now: Can we start after KubeCon if we consider this time as a break anyways? Specifically May 23rd?

jeremyrickard: If we start on May 23rd, then ~15 weeks later would be week of August 31st. 

A two week break between that and 1.26 means 1.26 would probably start around 19th of September. That would probably put us at the end of December if we held to 15 weeks. So I think we need to probably need to adjust a little bit either way? 12 weeks would end week of December 5th, 13 weeks would end week of December 12th, but also probably a lost week in there for KubeCon in October and a week for Thanksgiving in the US. Going much past the week of the 12th will run into a variety of holidays and end of the year.

releases/release-1.25/release-team.md
```diff
@@ -0,0 +1,17 @@
+# Kubernetes 1.25 Release Team
```
csantanapr, line 1: On numeral occasions, I'm looking for the team members table and I can't find it easily, just to realize that is in another file `release-team.md` and on the README.md.

Would be too much trouble to put the able at the end of the README.md instead of a separate file?

releases/release-1.25/README.md
jeremyrickard: > v1.26 is scheduled to release on Dec 20, 21, or 22 (15 week cycle) or Dec 27, 28, or 29 (16 week cycle to account for KubeCon NA).

I really think these dates are too late in December. Many people are going to be unavailable and I think it's a bit unfair to expect RT folks and RelEng folks to commit to being available during these timeframes....ESPECIALLY if we end up delaying at all. I think that Dec 13/14/15 would be the absolute latest we would want to release (imo). So if we bump out the start, I think we need to have a "short" release for that last one.

reylejano: @jeremyrickard I agree that any date after Dec 15 is too late

Although this PR is to set dates for the v1.25 release, I would also like to be cognizant of the v1.26 release.
The v1.25 release cycle can align with the Release Cadence KEP with the following dates:
- v1.25 starts on May 23
- v1.25 is scheduled to be released on Aug 30 for a 15 week cycle

Then the v1.26 release cycle may need a compromise like having a 1 week break after v1.25:
- v1.26 starts on Sept 6 (day after US Labor day holiday)
- v1.26 is scheduled for release on Dec 13, 14, or 15 (15 week cycle)

saschagrunert: @reylejano your latest proposal looks good to me!

releases/release-1.25/release-team.md
cici37: We have `links` at the top of readme file pointing to release team table: https://github.com/kubernetes/sig-release/blob/7d494b9a3bc717e6089b5ed5780b822c85abc841/releases/release-1.25/README.md?plain=1#L16

I hope that works^

releases/release-1.25/README.md
cici37: We certainly don't want to delay 1.26 too much. My concern is should we save some buffer for the release in case unexpected blocker coming up?  Out of this concern, I would prefer the current schedule of `1.25 starts on May 9th and released on 16th August for a 15 week cycle`. Otherwise we would have to stick to the schedule very strict in case of delaying 1.26.

saschagrunert: @cici37 which benefit do we gain over the first release week starting May 9th over May 23 when we consider KubeCon (or the "time around KubeCon") as non-development frame?

@reylejano can we shrink the release cycle to still start May 23 and end earlier, e.g. August 16 (14 week cycle)? This could be done by removing the alpha.3 release for example.

cici37: I assume it would depends on how we shrink it. We might end up losing one week of implementation or one week of releasing?(I don't think we could shrink one week of releasing time since we sure wanna things sink for some time. BTW, as @jeremyrickard mentioned earlier, there is no such thing like a "start date" of the enhancement reviewing, people could start the enhancement proposal and get it reviewed now for 1.25.

I am ok with 1.25 starts on May 23 as well.

reylejano: @saschagrunert @cici37 I think shrinking the release cycle to 14 weeks is fine
As Sascha suggested, burndown can start a week earlier at week 10, meaning that certain efforts associated with [burndown](https://github.com/kubernetes/sig-release/blob/master/releases/release_phases.md#burndown) will start a week earlier

cici37: @saschagrunert @helayoty @reylejano @jeremyrickard I have updated to a 14 weeks release starts at May 23rd. Please have a look when have time^

cici37:
I have updated the KEP with the proposal of `1.25 starts on May 23 and released on Aug 30 for a 15 week cycle`. Please feel free to raise concern of further suggestions^ Thank you

releases/release-1.25/README.md
cici37: Updated based on above discussion^

cici37:
Updated based on the discussion of having 1.25 starts on May 23 and released on Aug 23 for a 14 week cycle.

saschagrunert (approved):

PR #1931, opened 2022-06-08
Update roadmap

saschagrunert:
#### What type of PR is this:

/kind documentation

We now split-up the deliverables to reflect the state better (WIP vs TODO). This also means that we outline if no enhancement is available for a certain item.

We also move "Distribute the load of Kubernetes artifacts between vendors (Consumable)" to done since this is mostly in the scope of K8S infra now and we supported the initial phase.

We also move "Simplify CVE process for release management (Secure)" to done because there is just a single PR in the scope of Security Response Committee which has to be merged: https://github.com/kubernetes/committee-security-response/issues/63

None
Kudos to @LappleApple for the help on the new outline. :heart:

### DIFF
--- a/roadmap.md
+++ b/roadmap.md
@@ -30,11 +30,12 @@ publishing of Kubernetes related artifacts.
 ### Deliverables
 
 The following deliverables are necessary to achieve the overall goal. Within
-the following listing, all deliverables are sorted by their priority.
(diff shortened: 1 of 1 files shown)


cpanato (approved):
looks nice thanks!!
/lgtm

PR #1954, opened 2022-06-29
Remove duplicated GitHub token docs

ramrodo:
#### What type of PR is this:

/kind cleanup

This PR removes the GitHub token docs from the handbook since this specification was already moved to [krel docs](https://github.com/ramrodo/release/blob/master/docs/krel/release-notes.md#installation) where it makes more sense since there is where this token is used. See https://github.com/kubernetes/release/pull/2578.

Also, there is no other mention of this GitHub token in any part of this handbook.

None

None


cpanato (approved):
Thanks
/approve

PR #1958, opened 2022-07-07
[WIP] releng: Document process for Release Manager selection

justaugustus:
Signed-off-by: Stephen Augustus <foo@auggie.dev>

#### What type of PR is this:

/kind documentation

### DIFF
--- a/release-engineering/selection.md
+++ b/release-engineering/selection.md
@@ -1,0 +1,85 @@
+# Release Manager selection
+
+> DISCLAIMER: The initial draft of this document is intended to capture the
+> current state of the Release Manager selection process. It is neither a
+> perfect, nor complete process, but it is one that we have at present,
+> so hopefully documenting it is a step in the right direction.
+
+As described on the Kubernetes Community [website][release-managers-page], the
+Release Managers group is a group of Kubernetes contributors (under the Release
+Engineering subproject of SIG Release) which is tasked with publishing
+Kubernetes release artifacts, as well as maintaining the processes and tooling
+to do so.
+
+- [Expectations](#expectations)
+- [Identification](#identification)
+- [Selection](#selection)
+- [Promotion](#promotion)
+
+## Expectations
+
+(From ["Becoming a Release Manager Associate"][becoming-a-release-manager-associate])
+
+Contributors can become Associates by demonstrating the following:
+
+- consistent participation, including 6-12 months of active release
+  engineering-related work
+- experience fulfilling a technical lead role on the Release Team during a
+  release cycle
+  - this experience provides a solid baseline for understanding how SIG Release
+    works overall—including our expectations regarding technical skills,
+    communications/responsiveness, and reliability
+- working on SIG Release items that improve our interactions with Testgrid,
+  cleaning up libraries, etc.
+  - these efforts require interacting and pairing with Release Managers and
+    Associates
+
+There is currently no one accepted process for identifying or selecting
+candidate Release Manager Associates.
+
+## Identification
+
+In the past, we have done the following to consider candidates:
+
+- A candidate expresses interest to a Release Manager or SIG lead
+- A Release Manager or SIG lead recognizes the work of the candidate and
+  suggests they join the team
+- A SIG lead reaches out to current Release Team towards the end of the
+  release to seed candidates
+
+## Selection
+
+Once a candidate has been identified, SIG leads or senior Release Managers will
+do the following:
+
+- discuss the expectations of the role with them
+- confirm interest with the candidate
+- discuss the candidate with other Release Managers and SIG leads
+- vote on confirming the candidate as a Release Manager Associate
+
+If confirmed, members of Release Managers group will work with the candidate
+will work with the new Release Manager Associate on [onboarding][onboarding].
+
+## Promotion
+
+(From ["Becoming a Release Manager"][becoming-a-release-manager])
+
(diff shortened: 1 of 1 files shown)


release-engineering/selection.md
```diff
@@ -0,0 +1,85 @@
...
+- confirm interest with the candidate
+- discuss the candidate with other Release Managers and SIG leads
+- vote on confirming the candidate as a Release Manager Associate
+
+If confirmed, members of Release Managers group will work with the candidate
+will work with the new Release Manager Associate on [onboarding][onboarding].
+
+## Promotion
+
+(From ["Becoming a Release Manager"][becoming-a-release-manager])
```
puerco, line 65: OK, my $0.02 here is that we need to funnel the different stepping stones to the RM ladder and make them converge so that candidates start by shadowing a Branch Manager during one release. 

As we've been discussing, we want to open up the possibility of becoming a branch manager to more contributors, not just Release Team leads. This means that candidates will come from other realms and not just the RT but it also means that prior work we can consider as experience as release management experience may vary.

We can consider prior experience from many realms, but a tour as branch manager shadow should be required and should be what ultimately determines fitness for the role. Both giving candidates a real-world view of the RM duties and tooling while always having a guide to show them the way. On the opposite end, the tour through the release cycle will also show SIG Release if that person is willing to have the commitment.

In practice, this means dropping the "_experience fulfilling a technical lead role on the Release Team_" requirement in the [Becoming a Release Manager Associate](https://kubernetes.io/releases/release-managers/#becoming-a-release-manager-associate) section of the doc and coding the requirements to become an RMA as a first set of "soft", subjective requirements and then a "hard" one:

* "Soft": More subjective, consistent work, improving interactions with testgrid, etc
* "Hard": A cycle as Branch Manager shadow.

My opinion is that once both are met, collectively as a group, we can recognize the candidate and accept them as RMA.

saschagrunert:
I saw the PR, but will review once it got unWIPed.

justaugustus:
> I saw the PR, but will review once it got unWIPed.

@saschagrunert -- SGTM. Will give you a shout when this is ready!

---

@kubernetes/release-engineering -- For context, the [diagram](https://github.com/kubernetes/sig-release/pull/1958#discussion_r920277191) @puerco posted nicely captures a conversation that he and I had at Open Source Summit a few weeks back.

release-engineering/selection.md
```diff
@@ -0,0 +1,85 @@
...
+- demonstrating the willingness to lead
+- tag-teaming with Release Managers on patches, to eventually cut a release
+  independently
+  - because releases have a limiting function, we also consider substantial
+    contributions to image promotion and other core Release Engineering tasks
+- questioning how Associates work, suggesting improvements, gathering feedback,
+  and driving change
+- being reliable and responsive
+- leaning into advanced work that requires Release Manager-level access and
+  privileges to complete
```
jimangel, line 80: Without being a RMA one could meet most of those requirements. I would love to see more on how SIG Release can uplift RMAs. Is there any requirements for a Release Manager to even interact with a RMA? I do see that @puerco added in a branch manager shadow pre-req, but I do think there's more to SIG Release beyond just branch management.

It's possible I'm looking at RMA the wrong way, but I assumed the role was a a growth role that experienced leads / chairs could help mentor into RM, for example by:

- checking in with RMAs
- assigning issues
- pair-programming
- asking feedback
- sharing their thought process
- offloading that "one quick lgtm/approval" to the RMA and why (it takes 2 minutes vs. 30 seconds)

I by no means expect that level of mentorship all the time, but today, it is non-existent. I also want to be very clear, I don't want to discount the original list, I agree that independent work / drive and the above are all necessary.

Lastly, it's possible, with the TBD RMA period, my feedback goes away. There would be no need to uplift RMAs and there's an explicit prioritization of independent drive / collaborative work. If so, that should be explicitly stated that the RMA journey, that it is really just a trial for RM and there's no difference but permissions.

xmudrii: > Without being a RMA one could meet most of those requirements.

This is to some degree true, but for many tasks, you need elevated permissions (e.g. at least to read logs). There are some tasks that anyone can do, like Go updates and various other updates, but those tasks are not too common.

> Is there any requirements for a Release Manager to even interact with a RMA?

Yes. It's documented on the Release Managers page: https://kubernetes.io/releases/release-managers/#release-managers

From the linked page:

> Release Managers are responsible for:
> ...
> * Supporting Release Manager Associates and contributors through actively participating in the Buddy program
>   * Check in monthly with Associates and delegate tasks, empower them to cut releases, and mentor
>   * Being available to support Associates in onboarding new contributors e.g., answering questions and suggesting appropriate work for them to do

There are two major problems with this though:

* Availability of Release Managers in 2021/2022 has been a bit spotty (and that's totally okay, stuff happens!)
* To my knowledge, we never really had a streamlined process for this. For example, who is mentoring who. So far, it mostly works that RM brings some tasks on Slack and waits for someone to eventually pick them up. Now, because of the first point, we were running a little bit low on tasks, so there's that too

> It's possible I'm looking at RMA the wrong way, but I assumed the role was a a growth role that experienced leads / chairs could help mentor into RM, for example by:

I think that's a solid list of tasks and that's mostly how is it supposed to work. Maybe it would be nice to bring this up for discussion at some SIG Release meeting.

xmudrii:
/remove-lifecycle stale
@puerco @justaugustus Is there any help needed with this one?

PR #1967, opened 2022-07-14
Add outreach to SIGs through Slack during Week 1-2

rhockenbury:
#### What type of PR is this:
/kind documentation

This PR adds additional documentation around an outreach process that we've used on the enhancements subteam to be more proactive in the communication with SIGS regarding their enhancements.

### DIFF
--- a/release-team/role-handbooks/enhancements/README.md
+++ b/release-team/role-handbooks/enhancements/README.md
@@ -140,8 +140,14 @@ It is important that this process be followed and documentation remain up-to-dat
 - Mark features as `At Risk` if there is no communication, active PRs on the issues, or it is missing other requirements coming into Enhancement Freeze.
 - Start syncing with Communications Team on giving an induction what's coming up for the release.
 - Reach out to each [SIG on Slack](https://github.com/kubernetes/community/blob/master/sig-list.md) tagging one or more chairs or technical leads to communicate the Enhancements statuses
-  - For participating SIGs, mention the total number of Enhancements on the tracking spreadsheet for that SIG and the number of Enhancements that are `At Risk`
-  - For non-participating SIGs, mention the process and deadline for adding Enhancements to the `KEP Collection` tab on the tracking spreadsheet
+  - For participating SIGs, mention the total number of Enhancements on the tracking spreadsheet for that SIG and the number of Enhancements that are `At Risk`.  An example outreach on Slack could be structured as follows:
+
+    >    Hello sig-{{ SIG_NAME }}. Enhancements team here.  Just checking in as we approach enhancements freeze at {{ FREEZE_DATETIME }}. Your SIG has submitted {{ X }} enhancements for the {{ CYCLE }} cycle, and {{ Y }} enhancements are currently `At Risk`. Refer to the [announcement here](https://groups.google.com/a/kubernetes.io/g/dev/c/slwAhtVqjXk/m/zz-eeXdFBgAJ) for the list of review requirements. If your SIG still plans to submit any more enhancement, follow the [instructions here](https://groups.google.com/a/kubernetes.io/g/dev/c/slwAhtVqjXk/m/zz-eeXdFBgAJ) so the enhancements team can begin tracking. Please plan to make KEP updates to meet all of the requirements before enhancement freeze. Please reach out to [#release-enhancements](https://kubernetes.slack.com/archives/C02BY55KV7E) if you have any questions.
+
+  - For non-participating SIGs, mention the process and deadline for adding Enhancements to the `KEP Collection` tab on the tracking spreadsheet.  An example outreach on Slack could be structured as follows:
+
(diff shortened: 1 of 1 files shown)


rhockenbury:
@Priyankasaggu11929 - Open to thoughts on this.  Felt like it belongs in the rolebook since we've started doing this on the past 2-3 (maybe more?) releases and it's been fairly effective.

Priyankasaggu11929:
+1 for adding the "slack outreach" bit in the handbook.

Can we also include text stating that we want to tag one of each SIG chair and TL in order to draw their attention to our Slack messages?

From my experience, our slack outreach messages are generally lost in the SIG channels and receive little attention because they are sent by (mainly) new contributors, and we then hear the same issues later on that folks were not informed of the process/timelines/other-stuff, etc.

We can get contact information for SIG chairs/TL from their k/community/SIG README>, like - https://github.com/kubernetes/community/blob/master/sig-api-machinery/README.md

rhockenbury:
Yeah, it would more effective if we tagged the chairs / tech leads.  The chairs are also listed on the [sig list](https://github.com/kubernetes/community/blob/master/sig-list.md).  I'll can add some text around that - do you feel it's necessary to say "tag one chair and one technical lead" or keep it more general like "tag one or more chairs or technical leads"?

release-team/role-handbooks/enhancements/README.md
```diff
@@ -139,13 +139,16 @@ It is important that this process be followed and documentation remain up-to-dat
 - Stay on top of comments in issues when owners respond and update their status in the sheet if necessary.
 - Mark features as `At Risk` if there is no communication, active PRs on the issues, or it is missing other requirements coming into Enhancement Freeze.
 - Start syncing with Communications Team on giving an induction what's coming up for the release.
+- Reach out to each [SIG on Slack](https://github.com/kubernetes/community/blob/master/sig-list.md) to communicate the Enhancements statuses
+  - For participating SIGs, mention the total number of Enhancements on the tracking spreadsheet for that SIG and the number of Enhancements that are `At Risk`
+  - For non-participating SIGs, mention the process and deadline for adding Enhancements to the `KEP Collection` tab on the tracking spreadsheet
```
palnabarun, line 144: Why not email the SIGs on their respective mailing lists as well?

rhockenbury: There is an outreach via email to the k8s-dev group (mentioned on L142/L145) that has a summary of the number of KEPs per SIG. [See here ](https://groups.google.com/a/kubernetes.io/g/dev/c/NP3ByYMyWDU) for the email Priyanka sent for v1.25.  

I wouldn't be opposed to using the SIG mailing lists - presumably, there would be a better response rate than just sending a blanket email to k8s-dev. I also really like [Priyanka's suggestion](https://github.com/kubernetes/sig-release/pull/1967#issuecomment-1184004725) tagging the chair and technical lead on the Slack outreach to get a better response rate. 

Our strategy could be a combination of these approaches:
1. Reach out on Slack tagging chair/TL (~1 week prior to freeze) 
2. Reach out through SIG email list (~3-4 days prior to freeze if no response from 1)
3. Slack DM to chair/TL (~1 day prior to freeze if no response from 2)

However, for simplicity - I would opt to just continue with the Slack SIG outreach for the next few releases.  I think it's been effective enough and we don't want to overwhelm the enhancements team with too much additional work. What are your thoughts @Priyankasaggu11929?

Priyankasaggu11929: +1 to Ryler's comment. From past few release cycles, we've been doing several rounds of email outreach prior to the Enhancements Freeze.

However, we had to add the slack outreach part in the past couple of releases because people seemed to be more responsive there comparatively and it worked better to get actions/responses in time.

>I would opt to just continue with the Slack SIG outreach for the next few releases. I think it's been effective enough and we don't want to overwhelm the enhancements team with too much additional work.

I completely agree with this. Tracking enhancements most times becomes a lot work in itself - so, want to add as few layers of "outside github" outreach as possible to keep the team's bandwidth available.


---

Ryler, on the comment prior:

> do you feel it's necessary to say "tag one chair and one technical lead" or keep it more general like "tag one or more chairs or technical leads"?

general text like "tag one or more chairs or technical leaders" would work. Thank you! :)

gracenng: +1 on linking the [sig-list](https://github.com/kubernetes/community/blob/master/sig-list.md) in there for faster outreach.

Emailing the SIG email list imo is not doing more than what the Slack does and could be spammy.

Thoughts on adding the Slack outreach template in there so the future enhancements could just copy-paste? Small add to reduce friction

rhockenbury: Thanks Grace.  I do have the [sig-list linked](https://github.com/kubernetes/sig-release/pull/1967/files#diff-c853db51181791969e437e9f720a7ef21fbcf460f1eebda80b8ae6a7e2673ba8R142).   I've added the templates - I went with a blockquote rather than a code block cause I think it formats a little nicer in the rendered markdown.

gracenng:
Lovely! Thank you for the changes Ryler

/lgtm

Priyankasaggu11929:
Thank you for working on improving the handbook, Ryler. :)

PR #2002, opened 2022-08-24
Add exceptions.yaml for release-1.25

Priyankasaggu11929:
#### What type of PR is this:
/kind documentation

The PR adds the list of Exception requests received following the Enhancements Freeze and Code Freeze during the 1.25 release cycle.

cc: @cici37


cici37:
/lgtm
/approve
Thank you!
One more question: What is the purpose of listing the exceptions? If there is a same need for those reversion happening past freeze?

PR #2027, opened 2022-09-16
Remove deprecated calendar from RT lead handbook

mrbobbytables:
The calendar referenced below as the community calendar is not in use (hasn't been for a few years). Meeting invites that are sent to SIG lists and kdev will automatically be added to the community calendar listed here: k8s.dev/calendar

that old cal we have no control over, its not community owned and we can't take it down or do anything with it =/ we purged it from k/community some time ago but missed the references here.


palnabarun (approved):
/lgtm

PR #2205, opened 2023-04-02
Delay 1.27.0-rc.1 release to Thursday April 6th

xmudrii:
#### What type of PR is this:

/kind documentation

We should consider delaying the upcoming 1.27.0-rc.1 release to be able to pick up Go 1.20.3 update. This Go update is supposed to fix 4 CVEs in the standard library.

Some notes:

* We might want to wait for the announcement to go out and then verify if we are affected
  * However, verifying if we are affected tends to be complicated because Kubernetes is a huge project
* I decided to go with Thursday to give us enough time to update because it tends to take up to one day for Go images to become available
* I left the 1.27.0 release date as it is, hoping that there should be enough time from Thursday to Tuesday for downstream consumers to verify rc.1

xref https://github.com/kubernetes/release/issues/2991

/assign @salaxander @cici37 @jeremyrickard 
/hold for discussion


cici37:
/lgtm
/approve
Thanks for raising the PR!

palnabarun (approved):
/lgtm

saschagrunert (approved):

salaxander:
Thanks @xmudrii!

/unhold

PR #2229, opened 2023-05-02
clarify existing policy: release blocking jobs must only depend on community resources

BenTheElder:
#### What type of PR is this:

/kind documentation

AFAICT this has been the defacto policy for some time now, however we've not clearly stated it here.

This is an important requirement that avoids serious surprise issue like https://github.com/kubernetes/test-infra/issues/10043 (where we suddenly ran out of AWS quota pre-SIG-K8s-infra credits management and had to pull kops-AWS from kubernetes PR and release blocking jobs suddenly mid-release and never recovered this test coverage)

/sig release
/sig k8s-infra
/sig testing

### DIFF
--- a/release-blocking-jobs.md
+++ b/release-blocking-jobs.md
@@ -47,6 +47,7 @@ meet the below criteria completely due to technical debt.  Blocking jobs must:
 - Run at least every 3 hours
 - Be able to pass 3 times in a row against the same commit
 - Be Owned by a SIG, or other team, that is responsive to addressing failures, and whose alert email is configured in the job.
+- Depend on accounts and resources owned by The Kubernetes Project through SIG K8s Infra, so the project has visibility into funding and management of resources
 - Have passed 75% of all of its runs in a week, and have failed for no more than 10 runs in a row
 
 *In the case of failures, there must be an issue in kubernetes/kubernetes


jackfrancis:
Thanks @BenTheElder, this additional clarity is helpful

release-blocking-jobs.md
```diff
@@ -47,6 +47,7 @@ meet the below criteria completely due to technical debt.  Blocking jobs must:
 - Run at least every 3 hours
 - Be able to pass 3 times in a row against the same commit
 - Be Owned by a SIG, or other team, that is responsive to addressing failures, and whose alert email is configured in the job.
+- Depend on accounts resources owned by a SIG through SIG K8s Infra, so the project has visibility into funding and management of resources
```
xmudrii, line 50: Potential improvement/typo:
```suggestion
- Depend on accounts and resources owned by a SIG through SIG K8s Infra, so the project has visibility into funding and management of resources
```

xmudrii (approved):
/lgtm

release-blocking-jobs.md
BenTheElder: good catch, I also clarified "owned by a SIG through SIG K8s Infra" to "owned by The Kubernetes Project though SIG K8s Infra"

saschagrunert (approved):
LGTM, we still have to draw a line when we consider something as community owned, but this can be a case by case decision.

xmudrii (approved):

PR #2230, opened 2023-05-02
Document PR blocking job requirements

BenTheElder:
#### What type of PR is this:

/kind documentatino

The current status quo is that SIG Testing owners of Prow config have informally enforced this every since the introduction of the "release blocking jobs" document, but we've not had a good place to document it. We've been over this many times in slack, github issues, etc., but never captured it properly.

The obvious move is to just append this detail to the release blocking jobs document.

PR blocking jobs already have all of the requirements of release-blocking and the additional requirement that there be high-value in blocking PRs instead of only blocking releases (due to frequent breakage and the high cost of keeping the release blocking signal green without PR coverage, if such a cost exists).

Builds on https://github.com/kubernetes/sig-release/pull/2229

/sig testing
/sig release
/sig k8s-infra

### DIFF
--- a/release-blocking-jobs.md
+++ b/release-blocking-jobs.md
@@ -155,3 +155,23 @@ The general process to promote or demote jobs follows these steps:
 
 1. Open a kubernetes/test-infra PR that promotes or demotes the job to/from any
    of the sig-release boards.
+
+## Pull Request Blocking Jobs
+
+Blocking contributor PRs from merging is a serious consideration 
+and running a job on every PR push is expensive and potentially disruptive for
+contributors.
+Only after a job has met release blocking status AND shown to frequently go red 
+from bad code changes in Kubernetes should a job also be added to required on 
+all Kubernetes Pull Requests.
+
+When doing so you should contact SIG Testing and Release leads to discuss and
+provide evidence of the issues we've root caused that should've been caught prior
+to merge, and why existing presubmit coverage is not sufficient including new
+test cases within existing PR-blocking jobs (e.g. new unit and integreation tests)
+to regression test these bugs.
+
+If we catch a breakage in release-blocking but not PR blocking at most
+a few times per release, release-blocking is working as intended.
+Only when something is so poorly covered by presubmit that it breaks frequently
+should we escalate to requiring another PR-blocking job.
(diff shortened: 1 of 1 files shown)


BenTheElder:
/hold
for #2229 and discussion

BenTheElder:
aside: this seems like it should live in k/community (see e.g. https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/vendor.md) or k8s.dev but not a high-prio bikeshed :-)

release-blocking-jobs.md
```diff
@@ -154,3 +155,23 @@ The general process to promote or demote jobs follows these steps:
 
 1. Open a kubernetes/test-infra PR that promotes or demotes the job to/from any
    of the sig-release boards.
+
+## Pull Request Blocking Jobs
```
aojea, line 159: I remember a doc from Aaron with the criteria, so we only have only a subset of selected job, that run in less than one hour, high stability required, ...

aojea: found it https://github.com/kubernetes/test-infra/issues/6380#issuecomment-887158532

Pasting for reference

```
Let's lift loosely from my https://github.com/kubernetes/sig-release/issues/24#issuecomment-344381137 and https://github.com/kubernetes/sig-release/pull/346 that turned into what [release-blocking criteria are today](https://github.com/kubernetes/sig-release/blob/master/release-blocking-jobs.md#release-blocking-criteria-and-dashboard)

Looking at the merge-blocking jobs we have today, I would hope they meet these criteria:

The job has a 90th percentile duration of 75min (ideally <60min)
The job is better than a coin flip over all PR's (<50% daily failure rate across all PRs)
The job is better than four coin flips against your PR (95% consistent aka 5% flake rate)
The job is supported by contributors who can bring it back to green within a business day if it starts failing continuously or very near so
If the job is e2e, it must use OSS to provision the cluster (no hosted solutions)
```

aojea: can you incorporate that @BenTheElder ^^^ ?

BenTheElder: Ooh, thanks for digging that up!

I agree we should include more rigorous criteria and this seems like a good data-driven starting point.

BenTheElder: /hold
I'm going to come back to ping the thread again when I've incorporate this properly, for now going to get in #2229 first

aojea: ping @BenTheElder can we put a closure on this?

BenTheElder: Please fork the PR if you have bandwidth, still too many things going on.

PR #2240, opened 2023-05-10
Updating Enhancements Role-handbook

Atharva-Shinde:
Updates on enhancements tracking process

### DIFF
--- a/release-team/role-handbooks/enhancements/README.md
+++ b/release-team/role-handbooks/enhancements/README.md
@@ -90,6 +90,7 @@ Join the following Kubernetes Slack channels:
 
 - [#sig-release](https://kubernetes.slack.com/messages/sig-release)
 - [#enhancements](https://kubernetes.slack.com/messages/enhancements)
+- [#release-enhancements](https://kubernetes.slack.com/messages/release-enhancements)
 - (optional) [#prod-readiness](https://kubernetes.slack.com/messages/prod-readiness)
 - (optional) [#release-management](https://kubernetes.slack.com/messages/release-management)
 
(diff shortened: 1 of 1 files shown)


release-team/role-handbooks/enhancements/README.md
```diff
@@ -196,6 +197,10 @@ It is important that this process be followed and documentation remain up-to-dat
 - Stay on top of issues and continually monitor them twice a week and look at attached PRs. As Code Freeze gets closer, if there are PRs that have not been merged, move the issue to `At Risk`. If there is no activity, ping issue owners on either the issue or the k/k PR.
 - Monitor issues that are `At Risk` closely, almost daily. Code Freeze means no new code and keeping tabs on the status of the k/k PR is critical to planning. Make decisions if the enhancement should be deferred and work with SIG Leads to determine the best path forward.
 
+#### Before Code Freeze
+
+All enhancements going into the release must have their code pull requests in a merge-ready state with approved and lgtm labels applied. This includes tests, and docs PRs opened by the due date. Due to the Prow queue, this may mean that approved PRs merge shortly after the deadline.
```
marosset, line 202: Should we just link to this section so there won't be skew if one gets updated?
https://github.com/kubernetes/sig-release/blob/master/releases/release_phases.md#code-freeze

Atharva-Shinde: I definitely think it has its benefits but I'm in favour of keeping its uniformity with respect to other sections just for the sake of uniformity. Nevertheless the role-handbook will get updated every now and then. Wdyt?

Atharva-Shinde: Hey @marosset any inputs? Should I go on and make changes to include just the link?

marosset: I think we should at least link to the other link I mentioned. It is OK to restate the requirements but we should make it clear which one is the 'source of truth'

Atharva-Shinde: I agree! Thanks I'll add the link

gracenng:
/approve
Thanks for keeping this updated!!

PR #2246, opened 2023-05-16
Add experience to diversity list in team selection

bradmccoydev:
#### What type of PR is this:

/kind documentation

This PR will add experience to the diversity list in the release team selection to raise awareness for when leads select people that this is a mentorship program and we should also strive to have different levels of experience throughout the teams as discussed in a Slack thread of the team leads.

### DIFF
--- a/release-team/release-team-selection.md
+++ b/release-team/release-team-selection.md
@@ -39,6 +39,7 @@ Most importantly, strive for diversity in:
 - nationality
 - locality
 - company affiliation
+- experience
 
 ## Selection Criteria
 


leonardpahlke (commented):
/lgtm

Thanks for the addition. 

I think we have considered this in the past, that the RT shadow program is also generally an open source onboarding program for inexperienced people. We want to have space for new people. Also, some people among the newcomers are well versed in the Cloud Native community, others (also some returnees) may not be yet.

release-team/release-team-selection.md
```diff
@@ -39,6 +39,7 @@ Most importantly, strive for diversity in:
 - nationality
 - locality
 - company affiliation
+- experience
```
leonardpahlke, line 42: Should we add a bit context to it? We mention in the paragraph before that we select based on experience.

> We want to have space for new people. Also, some people among the newcomers are well versed in the Cloud Native community, others (also some returnees) may not be yet.

Should we add something like this?

leonardpahlke:
cc @gracenng

release-team/release-team-selection.md
gracenng: I think above already mentioned
> Having two shadows with prior Release Team experience and two new shadows per Release Team role is not a requirement and is a suggestion.
Maybe can do something like "experience in Cloud Native community"?

leonardpahlke:
```suggestion
- experience in open source communities
```
thoughts?

sftim: I get that we don't only want experienced contributors, and that it's good to be inclusive, but:

do we want to explicitly **avoid** a situation where the team is diverse in every other aspect but everyone already has open source contribution experience?

leonardpahlke: > do we want to explicitly avoid a situation where the team is diverse in every other aspect but everyone already has open source contribution experience?

The opposite is the case. When we add "experience in open source communities" to the list, we make sure we have room for experienced open source people and those who are new to open source. Diverse open source experience in the release-team. Section: https://github.com/kubernetes/sig-release/blob/master/release-team/release-team-selection.md#membership

leonardpahlke: I am okay with adding it as is without specifying it as by me before [here](https://github.com/kubernetes/sig-release/pull/2246#discussion_r1195279029)

Verolop: I can confidently say that the ratio of unqualified candidates, with no relevant technical experience who also meet the rest of the diversity checkboxes is much lower than our concerns might lead us to believe. I don't have "hard, cold numbers", but for every "unqualified yet diverse" individual who dares to apply, there are a bunch of "qualified and diverse" individuals who don't even dare to apply.
Experience in OSS communities is sometimes a privilege that not every one has had the time to allocate for, but they might be technically qualified and willing to learn. I also believe that as mentors we should do a better job onboarding people, which has also been a recurring topic in these conversations 😊

leonardpahlke: I don't have numbers either, but I think we already do a good job engaging inexperienced people to the release team. I don't think we have a diversity problem. Adding experience to the list just completes it a little more, and I think that's good.

leonardpahlke: @gracenng @bradmccoydev thoughts on getting the PR merged?

PR #2302, opened 2023-08-08
Add `hugo.toml` changes to docs handbook

Rishit-dagli:
The docs lead might have to change the `hugo.toml` file as well while updating previous releases which was missing from the docs handbook.

/kind documentation


cpanato (approved):
/lgtm

PR #2342, opened 2023-09-17
Update Release Notes Handbook to remove website PR creation instructions

ashnehete:
#### What type of PR is this:

/kind documentation

This PR removes the sections that instruct members to update [kubernetes-sigs/release-notes](https://github.com/kubernetes-sigs/release-notes) with new release notes. The process has been automated with https://github.com/kubernetes/release/issues/1087 and is no longer needed in the handbook.

Fixes https://github.com/kubernetes/release/issues/1087

cc @saschagrunert

### DIFF
--- a/release-team/role-handbooks/release-notes/README.md
+++ b/release-team/role-handbooks/release-notes/README.md
@@ -178,6 +178,7 @@ To update an entry in this section the following steps must be performed:
 Begin running release-notes tool for the ongoing collection of release notes with the first alpha release, which has been cut directly after the latest minor.
 
 - Update the `release-notes-draft.md`
+- Verify release notes are available on (relnotes.k8s.io)[https://relnotes.k8s.io/]
 - Begin attending release team meetings
 - Informal intro meeting with release notes team to discuss contact information and logistics
 
--- a/release-team/role-handbooks/release-notes/README.md
+++ b/release-team/role-handbooks/release-notes/README.md
@@ -70,7 +70,7 @@ In the first week of the release cycle, the Release Notes Lead will organize an 
 
 #### Early and mid release cycle (weeks 1-8) ~1-5 hours/week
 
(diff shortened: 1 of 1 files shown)


release-team/role-handbooks/release-notes/README.md
```diff
@@ -179,8 +176,8 @@ To update an entry in this section the following steps must be performed:
 ### Week 1
 
 Begin running release-notes tool for the ongoing collection of release notes with the first alpha release, which has been cut directly after the latest minor.
-    - Update the `release-notes-draft.md`
-    - Update the website
```
saschagrunert, line 183: Do we want to verify at least that the notes are available on the website?

ashnehete: That's a good idea. Currently, `krel release` publishes artificates to GCS which should instantaneously be picked up by the website. So is this the correct spot to add that?

saschagrunert: Yes, I think so :+1:

fsmunoz:
Actually, from a `krel` perspective and just to be absolutely on the safe side: this means that the `krel release-notes --create-website-pr --fix --fork fsmunoz --list-v2 ...` command is now not needed, and the `--create-draft-pr` is enough since the result of it will be picked up by the website?

ashnehete:
> Actually, from a `krel` perspective and just to be absolutely on the safe side: this means that the `krel release-notes --create-website-pr --fix --fork fsmunoz --list-v2 ...` command is now not needed, and the `--create-draft-pr` is enough since the result of it will be picked up by the website?

That is correct. I have also added a deprecation to the flag `--create-website-pr` at https://github.com/kubernetes/release/pull/3277.

saschagrunert (approved):

PR #2390, opened 2023-12-11
Update roadmap for 2024 and beyond

saschagrunert:
#### What type of PR is this:

/kind documentation

Updating the roadmap based on the recent refinement sessions.

cc @kubernetes/release-engineering @kubernetes/sig-release-leads 

None

None

### DIFF
--- a/roadmap.md
+++ b/roadmap.md
@@ -1,27 +1,21 @@
 # Roadmap and Vision
 
-## SIG Release Roadmap for 2022 and beyond
+## SIG Release Roadmap for 2024 and beyond
 
-This document contains the SIG Release Roadmap for 2022 and beyond. More
+This document contains the SIG Release Roadmap for 2024 and beyond. More
 detailed information can be found on the corresponding project boards.
 
 ### Primary Focus
 
-Establish a **consumable**, **introspectable**, and **secure** supply chain for
-Kubernetes. As a supply chain we understand the defining, building and
-publishing of Kubernetes related artifacts.
+Provide a more **robust**, **fast** and **flexible** release pipeline for
+Kubernetes.
 
-1. **Consumable**: Improving the usability of artifacts by making their
-   consumption easier. This includes being process independent of vendor,
-   employer and individuals.
-1. **Introspectable**: It is clear for users at which point and how Kubernetes
-   artifacts are being built. This includes the documentation of all
-   deliverables as well as clarifying what we do not support. All official
-   release artifacts will be built by a hermetic process that is impervious to
-   human interference.
-1. **Secure**: The artifacts we produce are verified for their integrity. This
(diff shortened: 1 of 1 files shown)


roadmap.md
```diff
@@ -50,51 +43,26 @@ state.
...
-   Outcome: Being able to promote files as artifacts and using this mechanism
-   for Kubernetes releases.
-
-1. **Define and collect metrics about Kubernetes releases (Introspectable)**
+1. **Make Kubernetes releases more flexible**
 
-   https://github.com/kubernetes/sig-release/issues/1527
+   Outcome: Ensure we migrate away from gs://kubernetes-release, use a
+   community-owned Object Storage bucket for binaries releases, serve the
+   binaries published through a CDN service.
```
saschagrunert, line 50: @ameukam is it fine if we make this a bit less specific? This way we could integrate topics like:

- Allow creating kubernetes releases from (private) forks
- Add parameters to krel to support other destinations (buckets)

ameukam: That will increase the timeframe to implement this. 
We may end up in a situation where the `gs://kubernetes-release` bucket is unavailable for the release process.
Better to do this ASAP (in 1.30).

saschagrunert: Good, reverted back to the original state.

roadmap.md
```diff
@@ -31,15 +25,14 @@ state.
...
+   Outcome: Provide a fast and error-resistant workflow for artifact promotion
+   compliant with SLSA.
 
-   Enhancement: https://github.com/kubernetes/enhancements/issues/3027
+   Enhancement: was https://github.com/kubernetes/enhancements/issues/3027
 
-   Project board: https://github.com/orgs/kubernetes/projects/138
+   Project board: https://github.com/orgs/kubernetes/projects/171
 
 1. **Moving deb/rpm package builds to community infrastructure (Consumable)**
```
xmudrii, line 37: This is in fact done, so I think we should rephrase this goal. I would split it into two deliverables:

- Making deb and rpm package builds more robust
  - The remaining work that we have is very well described with the provided robustness definition, e.g. writing tests, enabling reproducible builds...
  - Outcome can be similar to the outcome of artifact validation, e.g. provide a fast and error-resistant workflow for building and publishing packages to the community-owned infrastructure
- Enable other Kubernetes subprojects to use our packages infrastructure
  - This is about flexibility and it should cover efforts like building a CLI, library, writing documentation, etc. We don't have a KEP for this, but we can reuse the existing project board for now

saschagrunert: @xmudrii considering current resources, should we move both into the `Todo` section?

saschagrunert: Updated the roadmap to reflect the current state.

saschagrunert:
@kubernetes/release-managers @kubernetes/sig-release PTAL again

PR #2492, opened 2024-04-25
Move Major Themes responsibility from Release Notes to Comms

katcosgrove:
#### What type of PR is this:

/kind cleanup
/kind documentation

The collection of Major Themes is currently handled by Release Notes, when it is more in line with the duties of the Comms team and most prominently consumed by the Comms team. To better reflect the actual behavior of the teams, I have moved the primary responsibility for collection of Major Themes to Comms, and put Release Notes in a supporting role for this duty.

None

### DIFF
--- a/release-team/role-handbooks/communications/README.md
+++ b/release-team/role-handbooks/communications/README.md
@@ -53,10 +53,19 @@ There is a channel on the Kubernetes Slack workspace, `#release-comms`, which is
 ## Release deliverables
 
 Throughout the release cycle, the main Communications deliverables include:
-* **Authoring the Kubernetes release blog.** The Communications team writes the release blog, which is the official announcement of the Kubernetes release.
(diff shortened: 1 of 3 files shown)


katcosgrove:
/assign @a-mccarthy

Since you are the incoming Comms lead, this will affect you. Thoughts?

a-mccarthy:
Sounds good to me. I think it makes sense to move over the responsibility to comms since we are the primary folks using this content for our release deliverables. 👍

ramrodo (approved):
I agree with this proposal.
Thanks, Kat for organizing this.
/lgtm

katcosgrove:
/unhold

Thanks y'all!

fsmunoz:
Late to the party, but completely in agreement with the proposal, this was one of the suggestions I had for the 1.29 retro since the current situation is an artefact of how things were done in the beginning (when Release Notes were the single source of updates for a release, before there was a release blog), this suggestions greatly clarifies things and puts them in line with current practices.

/lgtm

shopdrop1 (commented):
@

PR #2494, opened 2024-04-29
Initial v1.31 docs

neoaggelos:
#### What type of PR is this:

/kind documentation

Adding initial documentation for v1.30 release and to discuss the potential release timeline.

References #2493 

cc @Priyankasaggu11929 @gracenng @katcosgrove


releases/release-1.31/README.md
```diff
@@ -0,0 +1,90 @@
...
+* [This document](https://git.k8s.io/sig-release/releases/release-1.31/README.md)
+* [Release Team](https://github.com/kubernetes/sig-release/blob/master/releases/release-1.31/release-team.md)
+* [Meeting Minutes](https://bit.ly/k8s131-releasemtg) (members of [release-team@] receive meeting invites)
+* [v1.31 Release Calendar](https://bit.ly/k8s-release-cal)
+* Contact: [#sig-release](https://kubernetes.slack.com/archives/C2C40FMNF) on slack, [release-team](mailto://release-team@kubernetes.io) on e-mail
+* [Internal Contact Info](https://bit.ly/k8s131-contacts) (accessible only to members of [release-team@])
+
+#### Tracking docs
+
+* [Enhancements Tracking Board](https://bit.ly/k8s130-enhancements)                                 <!-- TODO: replace with 1.31 board -->
```
Priyankasaggu11929, line 25:
```suggestion
* [Enhancements Tracking Board](https://bit.ly/k8s131-enhancements)                                 <!-- TODO: replace with 1.31 board -->
```

Priyankasaggu11929 (changes requested):

releases/release-1.31/README.md
```diff
@@ -0,0 +1,90 @@
...
+* [Feature blog Tracking Board](https://github.com/orgs/kubernetes/projects/161/views/4)            <!-- TODO: replace with 1.31 board -->
+* [Bug Triage Tracking Board](https://bit.ly/k8s131-bugtriage)
+* [CI Signal Project Board & Issue Tracking](https://bit.ly/k8s131-cisignal)
+* [Retrospective Document](https://bit.ly/k8s131-retro)
+* [kubernetes/sig-release v1.30 milestone](https://github.com/kubernetes/website/milestone/87)      <!-- TODO: replace with 1.31 milestone -->
```
Priyankasaggu11929, line 30:
```suggestion
* [kubernetes/sig-release v1.31 milestone](https://github.com/kubernetes/sig-release/milestone/27)
```

Priyankasaggu11929:
cc: @kubernetes/sig-release-leads

leonardpahlke (commented):
Did we discuss the timeline already? We may want to wait a few days for feedback. 

cc @dims @liggitt

releases/release-1.31/README.md
```diff
@@ -0,0 +1,90 @@
...
+| Major Themes complete                                                                  | Release Notes Lead            | Tuesday 20th August 2024                                                                                          | week 13  |                                                        |
+| 1.31.0-rc.1 released                                                                   | Branch Manager                | Tuesday 20th August 2024                                                                                          | week 13  |                                                        |
+| 1.31.0-rc.2 released                                                                   | Branch Manager                | Thursday 29th August 2024                                                                                         | week 14  | [1.31-blocking], [master-blocking], [master-informing] |
```
saschagrunert, line 86: How about omitting `v1.31.0-rc.2` and shift everything into the last week of August? This would give v1.32 one more week, which is already a short one due to holidays.

dims: +1 We definitely need to pull things in by at least a week please. All the previous 2nd release of the year were 1.22, 1.25, 128 were in August - looking at https://endoflife.date/kubernetes. If plan for Sep and we slip by a week that will make it hard for the last release of the year (fewer weeks there).

liggitt: +1, mid-August is where we've targeted in the past

liggitt: for reference, here are [historical release targets since 1.19 when we moved to 3 releases a year](https://docs.google.com/spreadsheets/d/1uhB218bDPayreXNHDuEY5VApi0W680vyoQ3I_7Mm3dc/edit?resourcekey=0-65TjeMTqsMLNH5fj3qxXPg#gid=0)

The mid-year target release date has ranged from August 4th - August 25th, with an average of August 17th.

liggitt: > Alternatively can start a week earlier like we did in 1.28?

Yeah, I'm not sure why we need a month before we "start" 1.31 ... I'm definitely already reviewing designs and code is merging.

neoaggelos: I'm more than fine pulling everything 1 week before. And if we also skip v1.31.0-rc.2, that should save two weeks worth of time. Are we confident that `rc.0` and `rc.1` are enough? 

Looking for a confirmation and I can adjust the schedule accordingly.

liggitt: I'm in favor of moving the start of the cycle as early as possible ... like early-mid May. I think that reflects reality more accurately, 1.31-targeted work is already happening.

I'd probably also start with the target release date and work backwards for the end dates. I think we normally try to release on a T/W/Th, so August 13/14/15 as a possible release date? Then if we have to slip a ~week we're still easily in August and don't crowd 1.32.

Planning rc.0 / rc.1 / release, with at least a week between each seems plausible to me, and if rc.1 still has enough issues to warrant an rc.2, we can consider slipping O(days) to accommodate.

neoaggelos: i've moved start to 20th may and release to august 22nd. @Priyankasaggu11929 do you think there would be enough time to handle the shadow applications if we started on the week of 13th may? I will be mostly unavailable during week 1, but we could still start I think

katcosgrove: Moving the start of the release to 20 May with a 22 August target for release is reasonable to me. That gives us enough time to run the shadow application and selection process, with a nice gap for 1.32. Pulling it any earlier than that will cut into shadow selection, but week 0 and week 1 are both so light for most teams it's doable if necessary.

Priyankasaggu11929: > I've moved start to 20th may and release to august 22nd. @Priyankasaggu11929 do you think there would be enough time to handle the shadow applications if we started on the week of 13th may?

I can send 1.31 shadow application form, as soon as tomorrow (May 1), run it for for 2 weeks (till May 15), and then announce/inform all applicants by May 22 at max.

So, starting as early as August 13/14/15 should be fine IMO. We could then just use the first week to kick-off 1.31 enhancements opt-in, while we conclude selections.

For sending out shadow application, I need a start date & tentative target release date, to inform applicants of their required time commitments.

katcosgrove: Do you mean starting May 13? I'm okay with that schedule, Week 1 doesn't have work for shadows (usually).

katcosgrove:
Thanks for adjusting the start and end dates! I'm happy with this. 

/unhold
/lgtm

releases/release-1.31/README.md
dims: thanks @neoaggelos @Priyankasaggu11929 @katcosgrove

saschagrunert (approved):

PR #2507, opened 2024-05-15
Align a few 1.31 summary dates with timeline

reylejano:
When we were going over dates in the SIG Docs meeting we noticed the 1.31 Summary dates did not align with the timeline.
In the summary, the date stays the same for Test Freeze but it shows as week 12 not week 11 in the timeline and for the Docs Freeze, the summary had 30th July August while the timeline shows 30th July.

/assign @Priyankasaggu11929 @neoaggelos


saschagrunert (approved):

neoaggelos (commented):
late /lgtm

PR #2542, opened 2024-06-25
Update release notes draft to version v1.31.0-alpha.2

satyampsoni:
**What type of PR is this?**
/kind documentation

**What this PR does / why we need it**:
This PR updates the Release Notes Draft to k/k v1.31.0-alpha.2

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:
This is an automated PR generated from `krel The Kubernetes Release Toolbox`

PR #2547, opened 2024-07-02
Lint yaml workflow update

Checksumz:
#### What type of PR is this:
/kind feature


.github/workflows/yaml-lint-checker.yaml
```diff
@@ -0,0 +1,68 @@
...
+  # Allow manual triggering
+  workflow_dispatch: {}
+
+concurrency:
+  group: ${{ github.ref }}-${{ github.workflow }}
+  cancel-in-progress: true
+jobs:
+  yaml-lint-checker:
+    name: Release Notes Check Yaml 👀
+    runs-on: ubuntu-22.04
```
cpanato, line 16:
```suggestion
    runs-on: ubuntu-latest
```

.github/workflows/yaml-lint-checker.yaml
```diff
@@ -0,0 +1,68 @@
...
+      - releases/**/release-notes/**.yaml
+  # Allow manual triggering
+  workflow_dispatch: {}
+
+concurrency:
+  group: ${{ github.ref }}-${{ github.workflow }}
+  cancel-in-progress: true
+jobs:
+  yaml-lint-checker:
+    name: Release Notes Check Yaml 👀
```
cpanato, line 15: drop the `eye` emoji

.github/workflows/yaml-lint-checker.yaml
```diff
@@ -0,0 +1,68 @@
...
+    runs-on: ubuntu-22.04
+    if: ${{ !github.event.pull_request.draft }}
+    steps:
+      - name: Cancel Previous Actions
+        uses: styfle/cancel-workflow-action@0.9.1
+        with:
+          access_token: ${{ github.token }}
+      - name: Check out code
+        uses: actions/checkout@v4
```
cpanato, line 24: use git hashes instead of tags, they are mutable 

```
uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
```

apply this pattern to the other actions

.github/workflows/yaml-lint-checker.yaml
```diff
@@ -0,0 +1,68 @@
...
+        with:
+          fetch-depth: '0'
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.22.4'
```
cpanato, line 30:
```suggestion
          go-version: '1.22'
          check-latest: true 
```

.github/workflows/yaml-lint-checker.yaml
```diff
@@ -0,0 +1,68 @@
...
+    if: ${{ !github.event.pull_request.draft }}
+    steps:
+      - name: Cancel Previous Actions
+        uses: styfle/cancel-workflow-action@0.9.1
+        with:
+          access_token: ${{ github.token }}
+      - name: Check out code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: '0'
```
cpanato, line 26: do you need to fetch the entire history?

cpanato (changes requested):
thanks for working on this
a few comments

.github/workflows/yaml-lint-checker.yaml
```diff
@@ -0,0 +1,68 @@
...
+        run: go mod download
+      - name: Build go prog
+        run: go build -o validate_yaml .github/scripts/validate_yaml.go
+      - name: Validate YAML
+        id: validate_yaml
+        run: |
+          INVALID_FILES=""
+          # check diff for invalid yaml files
+          git diff --name-only ${{ github.event.pull_request.base.sha }} ${{ github.sha }} -- releases/ | grep -E '\.ya?ml$'
+          CHANGED_FILES=$(git diff --name-only $(git merge-base origin/master HEAD) -- releases/ | grep -E '\.ya?ml$' || true)
```
cpanato, line 41: for this part you maybe can use this action that get all you need: https://github.com/tj-actions/changed-files

.github/workflows/yaml-lint-checker.yaml
```diff
@@ -0,0 +1,68 @@
...
+          if [ -n "$CHANGED_FILES" ]; then
+          echo "Validating yaml files: \n${CHANGED_FILES}"
+          OUTPUT=$(./validate_yaml $CHANGED_FILES)
+            if [[ "$OUTPUT" == *"YAML Validation Failed"* ]]; then
+              # If there are invalid files, set the output variable and fail the job. GITHUB_ENV is a special file that 
+              # allows you to set env variables for subsequent steps in the same job
+              echo "invalid_files<<EOF" >> $GITHUB_ENV
+              echo "$OUTPUT" >> $GITHUB_ENV
+              echo "EOF" >> $GITHUB_ENV
```
cpanato, line 50: this works? not 100% sure that this works

cpanato (changes requested):
i think this should go to the release-actions repo we have and then re-use here, to avoid having code in this repo.

wdyt @saschagrunert @puerco @xmudrii ?

.github/scripts/validate_yaml.go
```diff
@@ -0,0 +1,51 @@
...
+		if result != "Valid" {
+			invalidFiles = append(invalidFiles, result)
+		}
+	}
+
+	if len(invalidFiles) > 0 {
+		fmt.Println("YAML Validation Failed:\n")
+		for _, errorMsg := range invalidFiles {
+			fmt.Println(errorMsg)
+			fmt.Println("\n" + strings.Repeat("-", 40) + "\n")
```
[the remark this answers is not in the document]
npolshakova: I think we want to print all the invalid files, and then exit after the loop.

.github/workflows/yaml-lint-checker.yaml
Checksumz: Verified on my local branch that this allows to create a multiline variable for use in the next step.  Will send a screenshot with echo later

.github/scripts/validate_yaml.go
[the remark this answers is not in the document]
Checksumz: As Nina mentioned, this is to gather all the output into the variable on failure

npolshakova:
> i think this should go to the release-actions repo we have and then re-use here, to avoid having code in this repo.
> 
> wdyt @saschagrunert @puerco @xmudrii ?

Are [these](https://github.com/kubernetes/release/tree/master/.github/workflows) the release actions workflows? We'd want to be able to run this on release notes PRs going into sig-release and block if there's any invalid yaml in the diff. That would prevent having to create a separate PR to unblock the release notes like this [one](https://github.com/kubernetes/sig-release/pull/2541).

.github/workflows/yaml-lint-checker.yaml
npolshakova: As we edit the release notes, we usually keep the history: https://github.com/kubernetes/sig-release/pull/2542 
image

But then before we merge into the sig-release repo, we clean up everything into a single commit. 

I thought `fetch-depth: 1` only fetches the latest commit? So unless we get the entire history, the workflow will only run on a single commit (which might be fine since the last commit will have the entire diff, but makes it a little harder to identify the invalid yaml during the release notes review).

.github/workflows/yaml-lint-checker.yaml
cloudmelon:
```suggestion
    name: Fixed Release Notes Check Yaml
```

.github/workflows/yaml-lint-checker.yaml
cloudmelon: Release action from today's release meeting for reference : https://github.com/kubernetes-sigs/release-actions

.github/workflows/yaml-lint-checker.yaml
Checksumz: cool done, do we normally just use latest than a specific version? Not clear because you mentioned using a specific commit on the

.github/workflows/yaml-lint-checker.yaml
Checksumz: done

.github/workflows/yaml-lint-checker.yaml
Checksumz: Done

.github/workflows/yaml-lint-checker.yaml
Checksumz: checking this

Checksumz:
@cpanato  i have implemented most of your suggestions apart from exit(0). 

@npolshakova  @cpanato  @saschagrunert  Just want to confirm if i need to close this PR and reopen it with https://github.com/kubernetes-sigs/release-actions and update it to be a reusable workflow ?

PR #2609, opened 2024-09-04
Update v.132 information with project board and milestone

fsmunoz:
Update the `README.md` with the missing project board and milestone information, now that the board was created.

cc @katcosgrove @gracenng @tjons


gracenng:
The Feature blog tracking board link is pointing to the Enhancements board right now. Can we get it to point to the feature blogs tab (https://github.com/orgs/kubernetes/projects/195/views/6).

Since the link is long maybe we can get a bitly like Enhancements board. Just note that if we change the order of the views on the board the bitly might break @tjons

saschagrunert (approved):

PR #2610, opened 2024-09-04
fix code freeze date in summary

danwinship:
1.32 README lists code freeze as being Nov 7th/8th in one place, and 19th/20th in another.

Given that both places say "week 9" (which matches the earlier date), and the "Kubernetes v1.32 Sneak Peak" email gave the 7th/8th date, I'm assuming that's correct.

/kind documentation
/assign @fsmunoz


fsmunoz:
Hello @danwinship , you're correct, thank you for the fix: this happened because the original Test Freeze date survived in the summary. 

/lgtm

PR #2668, opened 2024-11-25
Clarification for `milestone-maintainer` post Code-Freeze (1/2)

gracenng:
Folks have been adding milestone to their PRs post Code-Freeze, which they shouldn't do.
We're adding documentation to discourage that behaviour.

/kind documentation


saschagrunert (approved):
Thanks for clarifying this (again)!

sftim:
@gracenng when would this be OK to unhold?

gracenng:
wanted @katcosgrove to read it first but should be fine
/unhold

PR #2738, opened 2025-02-26
Add extra feature blog opt-in message template

rytswd:
#### What type of PR is this:

/kind documentation

Add some clarification how the template can be used, and also introduces a separate template which utilises GitHub style markdown for callouts


bradmccoydev (approved):
LGTM nice addition

PR #2845, opened 2025-08-26
Update release notes draft to version v1.34.0-rc.2

rashansmith:
**What type of PR is this?**
/kind documentation

**What this PR does / why we need it**:
This PR updates the Release Notes Draft to k/k v1.34.0-rc.2

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:
This is an automated PR generated from `krel The Kubernetes Release Toolbox`


releases/release-1.34/release-notes/maps/pr-132890-map.yaml
```diff
@@ -0,0 +1,4 @@
+pr: 132890
+releasenote:
+  text: 'NONE.'
```
wendy-ha18: Could we reach out to the author of PR https://github.com/kubernetes/kubernetes/pull/132890 to make sure the PR didn't need release note or the author forgot to include it? 
Cause if the PR doesn't need release note, this file can be removed out of maps and release-notes-draft.json/.md file.

releases/release-1.34/release-notes/maps/pr-131215-map.yaml
```diff
@@ -0,0 +1,7 @@
+pr: 131215
+releasenote:
+  text: |-
+    Removed "endpoint-controller" and "workload-leader-election" FlowSchemas from the default APF configuration.
+
+    Migrate the lock type used in the leader election in your workloads from configmapsleases/endpointsleases to leases.
```
wendy-ha18: This release note may missing the Action required part? this is the full release note of this PR: https://github.com/kubernetes/kubernetes/pull/131215
```
Removed "endpoint-controller" and "workload-leader-election" FlowSchemas from the default APF configuration.

Action required: migrate the lock type used in the leader election in your workloads from configmapsleases/endpointsleases to leases.
```

wendy-ha18: I think we can twist this to make it slightly clearer on action required part and change double quotation to correct annotation for variables. Something like:

```
Removed `endpoint-controller` and `workload-leader-election` FlowSchemas from the default APF configuration. 

Action required: workloads using leader election were required to migrate their lock type from `configmapsleases` or `endpointsleases` to `leases`.
```

Feel free to help me wording it to keep it close to original purpose of the author.

wendy-ha18:
:red_circle: It seems like the `Action required:....` couldn't been captured by krel in a few release notes, but this part is quite important for end-users.

According to this discussion from Release Announcement: https://github.com/kubernetes/website/pull/51877#discussion_r2277577623. (view screenshot).

image


The PR [#131837](https://github.com/kubernetes/kubernetes/pull/131837 ) should reserve the `Action required:....` part as it will breaking behaviours of end-users and shouldn't be missed. 
```
Static pods that reference API objects are now denied admission by the kubelet so that static pods would not be silently running even after the mirror pod creation fails.
ACTION REQUIRED: Prior to upgrade, ensure static pods are not referencing API objects such as ServiceAccounts, ConfigMaps, Secrets, ResourceClaims, CSIDrivers, PersistentVolumeClaims, or ClusterTrustBundles.
```

However the newest release-note-draft.md/json only include a part of it:
```
Static pods that reference API objects were denied admission by the kubelet so that static pods would not be silently running even after the mirror pod creation failed. (https://github.com/kubernetes/kubernetes/pull/131837, [@sreeram-venkitesh](https://github.com/sreeram-venkitesh)) [SIG Auth, Node and Testing]
```

Could you please also help to include the `Action required: ...` back to release note for this PR as well? :pray:

releases/release-1.34/release-notes/maps/pr-132890-map.yaml
rashansmith: I reached out. Do we have a limit on when to expect to hear back vs. moving forward.

wendy-ha18: Normally when we have time, I would properly suggest to wait for contributor to get back 1 or 2 days cause sometime they will busy with their own schedule. However, considering these factors:
- timing of final release cut today. 
- the purpose of PR [132890](#132890) was for improving internal test coverage (added a conversion test to check that when a ReplicaSet object is converted to a ReplicationController) only, not end-user visible change as [Does my pull request need a release note](https://github.com/kubernetes/community/blob/master/contributors/guide/release-notes.md#does-my-pull-request-need-a-release-note) guidance here. 
- this PR is kind cleanup 

I would suggest we remove this map.

cc @Vyom-Yadav for awareness.

Vyom-Yadav: Let's clean this up here first and later we can get this corrected in the PR.

rashansmith: Author confirmed release note is not needed.

wendy-ha18: Thanks @rashansmith , so it's good to remove this map

Vyom-Yadav:
I don't see the following PRs in the release notes:

- https://github.com/kubernetes/kubernetes/pull/133516
- https://github.com/kubernetes/kubernetes/pull/133570
- https://github.com/kubernetes/kubernetes/pull/133587
- https://github.com/kubernetes/kubernetes/pull/133604
- https://github.com/kubernetes/kubernetes/pull/133409

releases/release-1.34/release-notes/maps/pr-132890-map.yaml
Vyom-Yadav: Removed that note

PR #2867, opened 2025-09-15
PRR Freeze is now a hard deadline with exception request

katcosgrove:
#### What type of PR is this:

/kind documentation

Updates the PRR Freeze from a soft, unenforced deadline to a hard deadline with an exception process at the request of the production readiness subproject.

cc: @soltysh @drewhagen @kubernetes/sig-release-leads

### DIFF
--- a/releases/EXCEPTIONS.md
+++ b/releases/EXCEPTIONS.md
@@ -44,6 +44,9 @@ Email them to:
 
 [You should have *very high confidence* on the “additional time needed” number - we will not grant multiple exceptions for a enhancement. The same goes for over-estimating on the time needed. If the exception request is rejected on the basis that the asking time is too high, we will not re-evaluate if the asking time decreases.]
 
+Exceptions after PRR freeze:
+Due to the proximity to enhancements freeze, it is important to file an exception at your earliest convenience. You may file an exception request in advance if you know your KEP will miss the PRR freeze deadline. All requests will be reviewed and either approved or rejected as they come in.  The release team will be responsible for approving or rejecting exceptions based on the questions above.  
+
 Exceptions after enhancements freeze:
 If your issue was removed during the enhancements freeze please file an exception at your earliest convenience.  All requests will be reviewed and either approved or rejected as they come in.  The release team will be responsible for approving or rejecting exceptions based on the questions above.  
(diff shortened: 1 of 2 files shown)


releases/EXCEPTIONS.md
```diff
@@ -44,6 +44,9 @@ Email them to:
 
 [You should have *very high confidence* on the “additional time needed” number - we will not grant multiple exceptions for a enhancement. The same goes for over-estimating on the time needed. If the exception request is rejected on the basis that the asking time is too high, we will not re-evaluate if the asking time decreases.]
 
+Exceptions after PRR freeze:
+Due to the proximity to enhancements freeze, it is important to file an exception at your earliest convenience. You may file an exception request in advance if you know your KEP will miss the PRR freeze deadline. All requests will be reviewed and either approved or rejected as they come in.  The release team will be responsible for approving or rejecting exceptions based on the questions above.
```
jenshu, line 48: Should we define a hard deadline to submit PRR exception requests (e.g. X days after PRR freeze), and specify when the updates need to be done by? I'm afraid that if they come too close to enhancements freeze it will be like a domino effect and they'll have to submit an exception request for enhancements freeze too. Can one have both a PRR freeze and enhancements freeze exception request open at the same time?

soltysh: I'd vote to simplify where we can, so if you know you're late you should be able to apply for both. You can probably estimate and explicitly say that one needs X days for PRR, and Y days for enhancement.

drewhagen: I think for this release, we will accept a simplified submission of both a PRR and Enhancement Exception at the same time.

However, I will warn that if and when this is done, our team that reviews the exceptions are likely to perceive a higher risk which might raise the bar for justification. I can imagine though that needing exceptions for PRR and Enhancements at the same time is a thing. (@jenshu correct me if I'm wrong but I thought we had at least one that was the case for 1.34?)

soltysh: > However, I will warn that if and when this is done, our team that reviews the exceptions are likely to perceive a higher risk which might raise the bar for justification

That's totally reasonable.

Issue #2375, opened 2023-11-25
Updating documentation: Post 1.29 Enhancements Role Handbook Updates


npolshakova:
Umbrella issue to improve the [Enhancements Role Handbook](https://github.com/kubernetes/sig-release/tree/master/release-team/role-handbooks/enhancements) to clarify some things that have been points of confusion in this (and previous) releases.

- [ ] Add a reference to clarify what “PRR Freeze” means to handbook
       - Link[ this email](https://groups.google.com/a/kubernetes.io/g/dev/c/CQ33yPqp-H4/m/hHO-NaQiAQAJ) from PRR team in our Enhancements handbook (maybe in the outreach templates)
       - PRR freeze has always been a soft deadline due to a limited PRR reviewers handling multiple KEPs, making it tough to finish all reviews by the deadline
       - notify people in [#prod-readiness](https://kubernetes.slack.com/archives/CPNHUMN74) about KEPs waiting for PRR reviews since the enhancement freeze is coming up next week with project board filter with KEPs missing PRR assignee
- [ ] Add section to clarify test freeze in the enhancements handbook
       - What is the test freeze
       - Clarify code freeze checks
- [ ] Add clarification for KEPs with only bug fixes going into release are fine to mark as `tracked for code freeze` for the purposes of enhancements tracking.
- [ ] Add instructions to help Comms team compile the [major themes](https://github.com/kubernetes/sig-release/tree/master/release-team/role-handbooks/enhancements#responsibilities) to handbook
- [ ] Add guidelines for creating new Project Board from [template](https://github.com/orgs/kubernetes/projects) in Github
       - If we need additional automation that’s more declarative, Mark started this PR a while back that we can revisit: https://github.com/marosset/sig-release/pull/1
- [ ] Add instructions for enabling/disabling test-infra job that does enhancement collection based on lead-opted-in label
       - https://github.com/kubernetes/test-infra/pull/30528
       - https://github.com/kubernetes/test-infra/pull/30965
- [ ] Clarify code freeze cut offs in handbook (PRs must at least be approved by SIG reviewers and in the merge queue)
- [ ] Add example for [exceptions.yaml file](https://github.com/kubernetes/sig-release/blob/master/releases/release-1.28/exceptions.yaml) in the handbook

npolshakova:
Some of these were addressed in:
- https://github.com/kubernetes/sig-release/commit/f8f88200af90227794f9977577310891b7350dc9 
[...]

PR #2886, opened 2025-09-30
Add clarification for KEPs with only bug fixes going into release are fine to mark as tracked for code freeze for the purposes of enhancements tracking.

whtssub:
#### What type of PR is this:

/kind documentation

- Added clarification in the `Before Code Freeze` & `Code Freeze Templates` section

Addresses part of the umbrella issue https://github.com/kubernetes/sig-release/issues/2375 to improve the Enhancements Role Handbook

### DIFF
--- a/release-team/role-handbooks/enhancements/README.md
+++ b/release-team/role-handbooks/enhancements/README.md
@@ -301,6 +301,8 @@ See the sig-architecture Enhancements [KEP Template](https://github.com/kubernet
 
 **All** enhancements going into the release must have their pull requests in a merge-ready state with approved and lgtm labels applied. This includes tests, and docs PRs opened by the due date. Due to the Prow queue, this may mean that approved PRs merge shortly after the deadline.
 
+> **Note**: KEPs that contain bug fixes as part of their implementation work (even if the primary work is fixing existing issues rather than adding new features) are perfectly acceptable to mark as `Tracked for code freeze` for the purposes of enhancements tracking. The key requirement is that all code changes are properly reviewed and approved, regardless of whether they introduce new functionality or fix existing issues.
+
 To avoid delays and ensure everything is in place for Code Freeze: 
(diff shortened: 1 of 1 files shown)


sreeram-venkitesh:
CC: @jenshu

release-team/role-handbooks/enhancements/README.md
```diff
@@ -301,6 +302,8 @@ See the sig-architecture Enhancements [KEP Template](https://github.com/kubernet
 
 **All** enhancements going into the release must have their pull requests in a merge-ready state with approved and lgtm labels applied. This includes tests, and docs PRs opened by the due date. Due to the Prow queue, this may mean that approved PRs merge shortly after the deadline.
 
+> **Note**: KEPs that contain bug fixes as part of their implementation work (even if the primary work is fixing existing issues rather than adding new features) are perfectly acceptable to mark as `Tracked for code freeze` for the purposes of enhancements tracking. The key requirement is that all code changes are properly reviewed and approved, regardless of whether they introduce new functionality or fix existing issues.
```
jenshu: Just to clarify, do you mean that KEPs that *only* contain bug fixes should still be tracked? I think this goes against the messaging from https://groups.google.com/g/kubernetes-sig-release/c/LTzDa7OyBSA/m/IrKW973HAQAJ which says that bug fixes don't need to be tracked

rayandas: Yeah, I think changes that do not affect current behavior don’t require a KEP and we don't have to track it.

whtssub: Makes sense, thank you @jenshu for pointing this out. I have updated them accordingly.

jenshu:
/lgtm

@rayandas can you review as well?

rayandas:
Looks good to me.

PR #2955, opened 2026-02-06
PRR freeze: clarify milestone update requirement

jenshu:
#### What type of PR is this:

/kind documentation

Tiny change to clarify that for PRR freeze, the `milestone` update is only required if the stage has changed

N/A


dipesh-rawat (commented):
/lgtm

PR #2965, opened 2026-02-25
add exceptions.yaml for release 1.36 with enhancements exceptions

jmickey:
#### What type of PR is this:

/kind documentation

Adding release 1.36 exceptions.yaml with all enhancements freeze exception requests


dipesh-rawat (commented):
@jmickey Could we also add the PRR exception request details in this PR? Also noticed the dates look mismatched compared to what we have on the [board (here)](https://github.com/orgs/kubernetes/projects/241/views/7), could we align them?

jmickey:
@dipesh-rawat I've updated the dates.

rytswd:
We discussed offline how we would be looking at PRR exceptions in the future, and it would be a retro discussion item for the release team.

/lgtm
/approve

PR #3072, opened 2026-08-07
chore: ignore macOS .DS_Store files

BenTheElder:
#### What type of PR is this:

/kind cleanup

Ignores macOS files that shouldn't be checked in.
I ran into this while working on another more substantive PR (https://github.com/kubernetes/sig-release/pull/3073)

This PR was AI assisted because :effort: / testing. I reviewed it. This is a common pattern, most of our repos already have a gitignore including at least `.DS_store`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+# macOS
+.DS_Store


justaugustus (approved):
Thanks for the catch, Ben!

PR #3097, opened 2026-09-11
Link the release reference generation page from the Docs handbook

Caesarsage:
The handbook asks the Docs Lead to confirm generated docs readiness but does not
say where that process is documented.

Links https://kubernetes.io/docs/contribute/generate-ref-docs/release-generation/
from that step in the timeline and from the Tools list, so the Docs Lead can
check readiness without needing to ask.

Reverse of kubernetes/website#57160, which links this handbook.

/cc @sreeram-venkitesh @katcosgrove @fsmunoz @kubernetes/sig-docs-en-owners


natalisucks:
LGTM from SIG Docs, thank you @Caesarsage 🚀

